PR #103, opened 2015-01-26
aii-ks: ks-post-reboot should have proper unit on systemd 

### LINKED ISSUE #100: ks-post-install hangs after initial ncm-ncd --co spma during ks-post-install
Observed, but not yet debugged following issue:
when during the ks-post-install of EL7 you login with ssh, and the initial `ncm-ncd --co spma` is running; the `ncm-ncd` completes without problems, but never gets to the `ncm-ncd --co --all` step.

the `ks-post-install` script just hangs.  without the ssh login, everything goes smooth.
(the install is using the remote logger, so no real reason to login in the first place)

Remarks
 * not sure this is el7 specific, but i've never seen it with el6
 * could be due to some issues with the remote logger (bufferering issue?)

stdweird:
This is to avoid timeout of the service (and the cleanup that comes with it)
Fixes #100

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -1062,7 +1062,14 @@ sub kspostreboot_tail
       $config->getElement (EMAIL_SUCCESS)->getTree;
 
     print <<EOF;
-rm -f /etc/rc.d/rc3.d/S86ks-post-reboot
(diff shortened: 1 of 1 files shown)


jrha:
Tested?

stdweird:
yes

PR #106, opened 2015-02-17
Add support for locking down the bootloader

gombasg:


stdweird:
test this please

stdweird:
@gombasg does ncm-grub support this too? (if not, can you open an issue for it?) 
i consider any passwords set in kickstart as temporary, since the transport of ks files themself can't be that well protected, so any password set during kickstart should be changed later by the respective component.

gombasg:
@stdweird I plan to add support to ncm-grub as well, when I find the time.

jouvin:
Do we really need both the support in kickstart and in `ncm-grub` or would `ncm-grub` be enough?

BTW, we agreed on not modifying `ncm-grub` anymore without doing a rewrite. I thought there was an issue about this but I don't see it... May be worth opening an issue as suggested by @stdweird .

stdweird:
@jouvin yes, it's password hash in a public unprotected kickstart. i'd assume the first thing to do is change it (similar to the root password hash in ks; you should change it to something else with ncm-authconfig).

jouvin:
My point was : is there any real value to add a password via Kickstart if you can do it through `ncm-grub` (which is not the case today probably)?

jrha:
Yes, because it closes a window between install and second boot (after configuration) where grub could be unprotected.

PR #108, opened 2015-03-08
aii-web: FIX errors and add JSON profiles suport

samary:
* Some variables were not initialized
* add support for JSON profiles via GET variable

### DIFF
--- a/aii-core/src/main/perl/aii-web.cgi
+++ b/aii-core/src/main/perl/aii-web.cgi
@@ -11,11 +11,14 @@ my $boothd = "default";
 
 # Quattor profiles
 my $web_root = "/var/www/html";
-my $profiles_path="profiles";
+my $profiles_path = "profiles";
 my $profiles_dir = $web_root.'/'.$profiles_path;
 
 # PXE Linux directory
 my $pxelinux_dir = "/osinstall/nbp/pxelinux.cfg";
+
+# Profile prefix
+my $profile_prefix = "_profile";
 
 # Number of different boot types
 my %boot_type;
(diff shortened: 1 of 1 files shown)


aii-core/src/main/perl/aii-web.cgi
```diff
@@ -9,11 +9,14 @@
 # is missing the node will boot from the hard disk
 my $boothd = "default";
 
-# Quattor XML profiles
+# Quattor profiles
 my $web_root = "/var/www/html";
 my $profiles_path="profiles";
```
stdweird: spaces around `=`

aii-core/src/main/perl/aii-web.cgi
```diff
@@ -79,9 +82,14 @@ sub Initialize {
   # find all profiles in directory
   my $d;
 
-
-  push @profiles,map { s/\.xml$//; $_.=".".$server_domain; } sort(grep(/\.xml$/, readdir(DIR)));    
-
+  # read GET and read profiles by extension (GET ?json)
+  if ($ENV{'REQUEST_METHOD'} eq "GET" and $ENV{'QUERY_STRING'}=='json') {
```
stdweird: should be `$ENV{'REQUEST_METHOD'} and $ENV{'REQUEST_METHOD'} eq "GET" and $ENV{'QUERY_STRING'} and $ENV{'QUERY_STRING'} eq 'json'` to avoid isues with undefined ENV variables

aii-core/src/main/perl/aii-web.cgi
```diff
@@ -79,9 +82,14 @@ sub Initialize {
...
+    push @profiles,map { s/\.json$//; $_.=".".$server_domain; } sort(grep(/\.json$/, readdir(DIR)));
```
stdweird: factor out the common code (ie `json` or `xml`), and do the push only once. (this is too much copy/paste)

stdweird: also, how does this work with `json.gz`?

aii-core/src/main/perl/aii-web.cgi
```diff
@@ -9,11 +9,14 @@
...
 my $boothd = "default";
 
-# Quattor XML profiles
+# Quattor profiles
 my $web_root = "/var/www/html";
 my $profiles_path="profiles";
 my $profiles_dir = $web_root.'/'.$profiles_path;
 
+# PXE Linux directory
+my $pxelinux_dir = "/osinstall/nbp/pxelinux.cfg";
```
stdweird: is it used somewhere? (it's not in this diff)

aii-core/src/main/perl/aii-web.cgi
```diff
@@ -9,11 +9,14 @@
 # is missing the node will boot from the hard disk
 my $boothd = "default";
 
-# Quattor XML profiles
+# Quattor profiles
 my $web_root = "/var/www/html";
```
stdweird: ideally, these should be the same values as `aii-shellfe` uses

stdweird:
test this please

aii-core/src/main/perl/aii-web.cgi
samary: Ok, was not part of my diff.

aii-core/src/main/perl/aii-web.cgi
samary: it used in line 97, 126, 208 and 228 but was not defined.

aii-core/src/main/perl/aii-web.cgi
samary: I did it with Config::Simple but it adds a new dependency : perl-Config-Simple (see https://github.com/samary/aii/tree/read_cfg)

How do you suggest to do this in a cleaner way ?

stdweird: @samary ideally, the `package aii_shellfev2;` from `aii-shellfe` is a proper module, and we could make a fake application instance, which would then get automatic access to the configfiles and the default values.
but it would require a refactor of the `aii-shellfe`, and not everyone might agree with this idea.
so leave it for now i think. i won't object the current code :smile:

jrha:
LGTM

Issue #107, opened 2015-02-25
aii-opennebula: Update OpenNebula templates from configure hook


alvarosimon:
Right now ONE templates are generated from aii install hook. This is ok if we want to install and configure a new VM but if we want to update the templates or just generate templates without VM instantiation we should move these functions to be executed by configure hook instead install.

jrha:
Fixed by #112.

jrha:
@alvarosimon can you try to remember to include "fixes #nnn" (as documented [here](https://help.github.com/articles/closing-issues-via-commit-messages/)) in your PRs so the issues get closed automatically and I don't have to chase them down manually.

PR #112, opened 2015-04-10
aii-opennebula: Include new install hook

alvarosimon:
This PR includes a new configure hook:
 * #107  : Update OpenNebula templates from configure hook 

It was included a new hook to create/update Opennebula templates so now:
 * configure: create/update vm/vnet templates. It also generates new vm images.
 * install: instantiate/stop vm
 * remove: remove images/templates/vm

**NOTE:** This PR requires aii-pxelinux `Configure` hook to work. https://github.com/quattor/aii/pull/114


stdweird:
@alvarosimon no real remarks on the code. can you confirm you tested this on altaria?

alvarosimon:
@stdweird I didn't test this in altaria yet, I'm still have problems to generate the rpm in my laptop (it's strange because I don't have this problem generating the component package). I hope to find where is the issue soon.. if not I will use a VM to generate the aii rpm :smile:

alvarosimon:
Tested in altaria using the new pxelinux Configure hook

PR #113, opened 2015-04-11
AII opennebula: use cleaned up test methods

stdweird:


stdweird:
will be fine once we can use new build-tools which contain https://github.com/quattor/maven-tools/pull/48

jrha:
You can now use new build-tools which contain quattor/maven-tools#48

stdweird:
@jrha updated and ok

PR #117, opened 2015-04-29
aii-ks: Correct el7 variant from rhel7rc to el7

jouvin:
Matches the template library expectation! Old name `rhel7rc` clearly inappropriate...


jrha:
Working late?

stdweird:
test this please

jrha:
There is a template named `rhel7rc` in this repo, should that not also be renamed?

PR #121, opened 2015-05-19
Fix developer-info maven template in header

stdweird:
Generates clean release files with properly templated headers


stdweird:
https://github.com/quattor/release/issues/93
@jrha not a blocker for the release

PR #124, opened 2015-05-20
Allow reinstallation of node and check final status in oneliners

stdweird:
To reinstall we need to be able to run `aii-shellfe --remove $node --configure $node --install $node` ; and it has to happen in that order (e.g. the `remove` triggers a remove hook to cleanup e.g. SINDES or IPA state).

Also `status` is now processed last to be able to show the final state of operations e.g. `aii-shellfe --rescue nodeY --status nodeY`


stdweird:
This might have some side effects with current oneliners that schedule multiple commands for the smae node(s), but the current order makes no sense imho (`--remove` after `--configure`? `status` before commands that can change the state?)

samary:
Maybe we could add a clean `--reinstall` switch ?

stdweird:
@samary open an issue so we can discuss what `reinstall` means?

jrha:
LGTM, if we decide to implement reinstall we should have a new PR.

Issue #119, opened 2015-05-07
aii-opennebula: set AR network mac addr as optional 


alvarosimon:
Use opennebula generated mac addr if it is not defined by the hardware quattor templates

jrha:
Fixed by #127

PR #127, opened 2015-05-28
aii-opennebula: set VNET single address range mac addresses as optional

alvarosimon:
This PR includes a small fix to set AR MAC addr as an optional value, if it is not defined within hardware quattor tree the value is generated by OpenNebula:
http://docs.opennebula.org/4.12/user/virtual_resource_management/vgg.html#virtual-network-model
 * #119 : set AR network mac addr as optional 
 * #128 : prints that the VM was successfully created while there was an issue during instantiate.


aii-opennebula/src/main/resources/network_ar.tt
```diff
@@ -11,7 +11,9 @@ IP = "[% system.network.interfaces.${tmpbridge}.ip %]",
 [%         ELSE -%]
 IP = "[% system.network.interfaces.${pair.key}.ip %]",
 [%         END -%]
+[%         IF hardware.cards.nic.${pair.key}.hwaddr.defined -%]
```
stdweird: but this is not optional in the hardware schema, so will always be present. we  need to set a private MAC and somehow define that if this MAC is set, it is to be ignored in the ONE template.

aii-opennebula/src/main/resources/tests/profiles/vm.pan
```diff
@@ -97,7 +105,8 @@ bind "/system/opennebula" = opennebula_vmtemplate;
 prefix "/system/opennebula";
 "vnet" = nlist(
     "eth0", "br100",
-    "eth1", "br101");
+    "eth1", "br101",
+    "eth2", "br101");
```
stdweird: you can add a new attribute here for the mac to be ignored.
```pan
"ignoremac" = "somethign:private";
```

aii-opennebula/src/main/resources/network_ar.tt
stdweird: like 
```
[% IF hardware.cards.nic.${pair.key}.hwaddr == system.opennubula.ignoremac -%]
```

aii-opennebula/src/main/resources/vmtemplate.tt
```diff
@@ -22,7 +22,9 @@ IP = "[% system.network.interfaces.${tmpbridge}.ip %]",
 [%         ELSE -%]
 IP = "[% system.network.interfaces.${pair.key}.ip %]",
 [%         END -%]
+[%         IF hardware.cards.nic.${pair.key}.hwaddr.defined -%]
```
stdweird: same logic here (and if this is common to the above TT, maybe factor it out?)

aii-opennebula/src/main/resources/tests/profiles/vm.pan
alvarosimon: @stdweird ok I see, if the hardware mac is always present in quattor  what about include ignoremac as a boolean type to be used by AII configure hook?:
```pan
"ignoremac" ? boolean = false # use quattor hw mac address instead OpenNebula mac prefix
```
but this will affect all the VM ARs, what do you think?

stdweird: that's another option, but it will affect all ARs of the VM.

alvarosimon: yes with different pros and cons @samary @StephaneGerardVUB do you have any preference about these options? use an ignoremac option per interface or to use it as global boolean

aii-opennebula/src/test/resources/vm.pan
```diff
@@ -57,27 +58,38 @@ prefix "/system/network";
...
+    "eth1", "altaria.vsc",
+    "eth2", "altaria.vsc");
 
-"datastore" = nlist(
+"datastore" = dict(
     "vda", "ceph.altaria",
     "vdb", "ceph.altaria");
+
+"ignoremac" = list (
+    "eth2",
```
stdweird: also test for mac address

aii-opennebula/src/main/resources/tests/profiles/vm.pan
StephaneGerardVUB: ignoremac per interface looks better to me

alvarosimon: @StephaneGerardVUB I have talked with @stdweird about this, we will implement both, you can use network interfaces or mac addrs, so the users can choose the best option that suit their needs. Now i'm fixing the schema that's a bit ugly... :sweat_smile: I hope to finish today

alvarosimon:
@stdweird @StephaneGerardVUB remarks addressed

aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
```diff
@@ -86,7 +88,23 @@ type opennebula_vmtemplate_datastore = string{} with {
...
+        };
+    };
+    return(true);
+};
+
+
 type opennebula_vmtemplate = {
     "vnet" : opennebula_vmtemplate_vnet
     "datastore" : opennebula_vmtemplate_datastore
+    "ignoremacaddr" ? type_hwaddr[] # MAC address will be ignored for these mac addr type_hwaddr
```
stdweird: use a new type `opennebula_ignoremac` which has attributes `addr` and/or `interface`. it's cleaner then common prefixes in attribute names

alvarosimon:
Documentation and schema updated to split /system/opennebula/ignoremac in to for interfaces and mac addr

alvarosimon:
remarks addressed

samary:
Hi,

@stdweird , @alvarosimon ,

After thinking about it, I came up to the conclusion that we should set this parameter at the network interface level (hardware) and not specific to ONE : 

Since : 
* This is related to the hardware (real vs virtual)
* This can be needed in other use cases (OpenStack ?)
* The mac address can be consumed by other components (eg 'ncm-network'  : 'set_hw_addr')

IMHO, we should be able to achieve this just by extending the /hardware schema with ignore_mac and set a default to 'false'

eg:

'''
"cards/nic/eth0/hwaddr" = "PLACE_HOLDER";
"cards/nic/eth0/boot" = true;
"cards/nic/eth0/ignore_mac" = true;

"cards/nic/eth1/hwaddr" = "PLACE_HOLDER";
"cards/nic/eth1/boot" = false;
"cards/nic/eth1/ignore_mac" = true;
'''

And then, ask to the AII-hook and NCM-Network to consume this information with the correct options.

stdweird:
@samary then we should make the `hwaddr` optional, and fix aii/ncm-network.
whatever interacts with openstack should make the decision to use or ignore the macaddress of a vm. but this info should indeed be common, e.g. `/sysem/cloud` rather then current `'/system/opennebula`.

i remember there was some discussion about it during the workshop, don't remember what the conclusion was.

samary:
@stdweird 
I don't like having this /system/cloud.
I really think this should be part of the /hardware branch since this is the hardware that we want to change. 
Having the hw_addr optional could lead to some inconsistency (can be forget in case of classical machine). This issue can be avoided via 3 ways : 

* Keeping hw_addr required, but allow a defined "PLACE_HOLDER".
* Remove the 'ignore_mac', and if a placeholder is defined instead of a regular mac (regex against xx:xx:xx:xx) , then act like if ignore_mac was set to true (placeholder should be like : IGNORE_MAC or something like that).
* Set hw_addr optional only if ignore_mac=true

I'm in favor of the 2nd

jrha:
Any reply to @samary's comments?

alvarosimon:
for me any option is good :smile: , maybe it should be discussed by quattor gurus, @stdweird any thoughts?

stdweird:
i gave it some more thought:
a. if the cloudinfrastructure sets/generates/defines the macaddress, you cannot set `set_hwaddr` to `true` (either per interface or as default for all iterfaces via `/system/network/set_hwaddr`). if you do, your network will work when using ncm-network
b. the tool that consumes the VM profiles to create the VMs (in this case `aii-opennebula`) can act upon this as they please, although this can lead to issues when swicthing cloud stacks. so we need to standardise. so i propose:
 1. if `set_hwaddr` is `true` for the interface (either global or set for the specific interface), the tool has to use the macaddr specified
 2. if `set_hwaddr` is `false` for the interface, the tool cannot use the macaddr specified and has to generate one
  3. the default when `set_hwaddr` is not set (system wide and interface), i would follow ncm-network and do as if `set_hwaddr` is `false`

also note that without `set_hwaddr` and using ncm-network, you either have to stay with 1 interface or use dhcp to set the IPs (or feel lucky :smile: ). 

also, at least openebula uses a welldefined deterministic function to generate the MAC based on the IP assigned to the interface, you can also use that in quattor to set the hwaddr in the profile. 
@alvarosimon can you add this function in the pan code so people can use this if they want to?

alvarosimon:
@stdweird sure we can use the same code used by Opennebula to generate the MACs from the IPs

alvarosimon:
@samary just one more question, because after a while I'm not quite sure what you want to achieve and maybe we can simplify this a bit. 
If I understood correctly, the problem that you have is you want to use the MAC addresses generated by opennebula to set not only opennebula aii, you want  also to configure dhcp and other network services from quattor using those pregenerated MACs right?

samary:
@alvarosimon : this is right. We want to use the generated MAC to configure DHCP and everything related. But this lead to an discrepancy between the content of the profile (empty MAC) and the AII configuration (generated MAC). And also, NCM-Network won't be able to consume this generated MAC.

@stdweird : Is there any way to have this generated MAC in the profile ? May we implement the generation algorithm in PAN so the MAC is set correctly in the profile ?

stdweird:
@samary that's what i meant, but then you might as well force it like it does now instead of opennebula generating it (and maybe choose another algorithm to generate the mac address)
it was my impression that your greatest issue is that you don't want to create hardware profiles foreach VM due to the mac addresses.

samary:
@stdweird We want to define the hardware template the same way we use to do it with real hardware. The only difference I see is the MAC address we want to leave empty and generated at build time. The corresponding algorithm should be chosen depending on the value of /system/cloud/? (opennebula, openstack, ...). If we can implement this on build, this will solve all issues.

alvarosimon:
@samary ok I see, so as @stdweird said the best option is to generate those MACs as Opennebula does using a function in quattor. 
You only need to know a priori the VM IPs, that's all and fill your quattor hardware database with these values. Correct me if I'm wrong, but for quattor point of view it does not matter if you are using pregenerated or "real" MACs... everything else will work as usual. As @stdweird said we only need to include the function to generate those MACs. I can write it, the code is is available from opennebula github. Besides it's easy to identify the ONE macs because it always use a prefix available from oned.conf `MAC_PREFIX = "02:00"` by default.

samary:
@alvarosimon This would be fantastic ! This seems to be the best approach. 
I don't know if we can change the MAC prefix in opennebula (if yes, we should put this as default in the schema but overridable, if not, hardcoded is fine for me).

alvarosimon:
@samary yes you can change the MAC prefix if you want. By default opennebula component schema uses this value  `"mac_prefix" : string = '02:00'`. And we must use the same value from the function in the pan code to generate the new MACs.

alvarosimon:
So if you agree give me some time to write the funcion... :smile:

samary:
@alvarosimon Nice, waiting forward to see that :smile:

alvarosimon:
@samary @stdweird I think that is almost done but it requires a review. 
I have included new functions to replace the VM hwaddr using the opennebula style `(MAC_PREFIX:IP)`. So basically you only need to set `OPENNEBULA_AII_REPLACE_MAC` variable and the function will do the work for you. This way the new MAC can also be used by any Quattor component and not only by the opennebula aii.
I have tested it and its working but please let me know if you do agree with these changes...

samary:
@alvarosimon , Thanks. We will try to test this asap on our site.

aii-opennebula/src/main/pan/quattor/aii/opennebula/default.pan
```diff
@@ -1,10 +1,12 @@
 unique template quattor/aii/opennebula/default;
 
 include 'quattor/aii/opennebula/schema';
+include 'quattor/aii/opennebula/functions';
 
 #  undef are set via schema default
 variable OPENNEBULA_AII_FORCE ?= undef; 
 variable OPENNEBULA_AII_ONHOLD ?= undef;
+final variable MAC_PREFIX = '02:00';
```
stdweird: can't be final, must be `variable MAC_PREFIX ?= '02:00';`

aii-opennebula/src/main/pan/quattor/aii/opennebula/default.pan
```diff
@@ -58,3 +60,10 @@ bind "/system/aii/hooks" = nlist with validate_aii_opennebula_hooks('remove');
 };
 
 bind "/system/aii/hooks" = nlist with validate_aii_opennebula_hooks('post_reboot');
+
+# If required replace VM hwaddr using OpenNebula fashion
+"/hardware/cards/nic" = if (exists(OPENNEBULA_AII_REPLACE_MAC) && exists(MAC_PREFIX)) {
+                                        opennebula_replace_vm_mac(MAC_PREFIX);
```
stdweird: cleanup indentation

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+
+https://github.com/OpenNebula/one/blob/master/share/router/vmcontext.rb
+
+Syntax: 
+mac_prefix:string ipv4:string
+
+mac_prefix hex:hex value used also by oned.conf (02:00 by default)
+ipv4 IP used by the VM
+}
+function ip2mac = {
```
stdweird: the function name needs to have a reference to opennebula in it

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+    # Check arguments
+    if (ARGC != 2) {
+        error("usage: \"hwaddr\" = ip2mac(MAC_PREFIX, IP)");
+    };
+    # Sanity check
+    if (!match(ARGV[0],
+        '^([0-9a-f]{2}[:]){1}([0-9a-f]{2})$')) {
```
stdweird: clean this up (`^[0-9a-f]{2}[:][0-9a-f]{2}$` ?)

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+        error("Invalid MAC_PREFIX format ("+ARGV[0]+")");
```
stdweird: use `format`

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+    };
+    if (!is_ipv4(ARGV[1])) {
+        error("Invalid IPv4 format ("+ARGV[1]+")");
```
stdweird: use `format`

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
stdweird: the name should also reference the fact that this is ipv4 only

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+"/hardware/cards/nic" = opennebula_replace_vm_mac(MAC_PREFIX);
+}
+function opennebula_replace_vm_mac = {
+    # Check for arguments
+    if (ARGC != 1) {
+        error("usage: opennebula_replace_vm_mac(MAC_PREFIX)");
+    };
+    # Sanity check
+    if (!match(ARGV[0],
+        '^([0-9a-f]{2}[:]){1}([0-9a-f]{2})$')) {
```
stdweird: same mac regexp as above

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+        error("Invalid MAC_PREFIX format ("+ARGV[0]+")");
```
stdweird: use `format`

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+    };
+    foreach (ethk;ethv;value("/hardware/cards/nic")) {
+        if ((exists(ethv['hwaddr']))) {
+                hwaddr = ethv['hwaddr'];
```
stdweird: fix indentation

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+                eth = ethk;
+                foreach (interk;interv;value("/system/network/interfaces")) {
+                    if ((match(interk, eth))) {
```
stdweird: use `interk == eth` ?

aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
```diff
@@ -86,7 +88,28 @@ type opennebula_vmtemplate_datastore = string{} with {
...
+    if (! exists("/system/network/interfaces/"+SELF)) {
+         error(format("ignoremac.interface: '%s' is not available from /system/network/interfaces tree", SELF));
+         return(false);
+    };
+    return(true);
+};
+
+@documentation{ 
+Type that sets which net interfaces/MACs
+will not include MAC values within one templates
```
stdweird: `one`-> `ONE`

alvarosimon:
Remarks addressed

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+    };
+    if (!is_ipv4(ARGV[1])) {
+        error(format("Invalid IPv4 format: %s", ARGV[1]));
+    };
+    # Convert IP octets to Hex
+    ipoctets = list();
+    foreach (i; octet; split('\.',ARGV[1])) {
+        ipoctets[i] = format("%02x", to_long(octet));
+    };
+    return(format("%s:%s", ARGV[0], join(':', ipoctets)));
```
samary: Where does this 'join' function come from ? I cannot compile with PAN.

alvarosimon: @samary you are right, we use join function internally so I have replaced it by format function. Please let us know if it's working now

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+    };
+    if (!is_ipv4(ARGV[1])) {
+        error(format("Invalid IPv4 format: %s", ARGV[1]));
+    };
+    # Convert IP octets to Hex
+    ipoctets = list();
+    foreach (i; octet; split('\.',ARGV[1])) {
+        ipoctets[i] = format("%02x", to_long(octet));
+    };
+    return(format("%s:%s:%s:%s:%s", ARGV[0], ipoctets[0], ipoctets[1], ipoctets[2], ipoctets[3]));
```
samary: @alvarosimon Shouldn't be %s:%s:%s:%s:%s:%s ?
A mac address is composed by 6 pairs and not 5

Anyway, we managed to write our join function. IMHO, this kind of functions should be ported back to quattor because they could be useful.

samary: OK, I missed it. Sorry

aii-opennebula/src/main/pan/quattor/aii/opennebula/functions.pan
```diff
@@ -0,0 +1,81 @@
...
+    if (!match(ARGV[0],
+        '^[0-9a-f]{2}[:][0-9a-f]{2}$')) {
+        error(format("Invalid MAC_PREFIX format: %s", ARGV[0]));
+    };
+    if (!is_ipv4(ARGV[1])) {
+        error(format("Invalid IPv4 format: %s", ARGV[1]));
+    };
+    # Convert IP octets to Hex
+    ipoctets = list();
+    foreach (i; octet; split('\.',ARGV[1])) {
```
stdweird: how about
```pan
macaddr = ARGV[0];
foreach (i; octet; split('\.',ARGV[1])) {
    macaddr = format("%s:%02x", macaddr, to_long(octet))
}
```

aii-opennebula/src/main/pan/quattor/aii/opennebula/default.pan
```diff
@@ -58,3 +60,10 @@ bind "/system/aii/hooks" = nlist with validate_aii_opennebula_hooks('remove');
 };
 
 bind "/system/aii/hooks" = nlist with validate_aii_opennebula_hooks('post_reboot');
+
+# If required replace VM hwaddr using OpenNebula fashion
+"/hardware/cards/nic" = if (exists(OPENNEBULA_AII_REPLACE_MAC) && exists(MAC_PREFIX)) {
+                            opennebula_replace_vm_mac(MAC_PREFIX);
```
stdweird: indentation

alvarosimon:
@stdweird @samary remarks addressed

stdweird:
LGTM
@jrha can you do the honours?

PR #131, opened 2015-06-16
Support all five systemd device naming schemes

jrha:
As documented at:
  * http://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
  * https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/ch-Consistent_Network_Device_Naming.html

This is the AII counterpart to quattor/configuration-modules-core#528.


stdweird:
@jrha you are aware that in el7 you can set any device name you want via pxe kickstart options if you have static network (it gets translated in udev rules).

stdweird:
lgtm :+1: on the unicode regexp

jrha:
Do we care about wireless adapters? Because then I need to extend support to all character prefixes, see: http://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-net_id.c#n20

Issue #115, opened 2015-04-22
questions around wipe_metadata


kwaegema:
There are some things in the generated kickstart files that needs some cleanup:
- `wipe_metadata` only has 1 parameter, the device. But it is further called with an additional value , eg:
 `wipe_metadata /dev/sda1 1`
- the `if [ primary != "extended" ]` wrapped around `wipe_metadata` : Should `primary` be a string or a variable? don't know where this should come from.

stdweird:
to answer the questions: 
a. the second parameter is the `clear_mb`, the minimum amount of data to write (affected by offset and/or alignment)
b. fixed that bug

stdweird:
@kwaegema need anything else?

kwaegema:
Nope, that clears it up, thanks!

PR #133, opened 2015-08-06
aii-ks: improved disk/partiton removal in pre

stdweird:
Fixes https://github.com/quattor/ncm-lib-blockdevices/issues/50 and https://github.com/quattor/aii/issues/115


jrha:
LGTM

PR #135, opened 2015-08-16
aii-shellfe: add --reinstall option

### LINKED ISSUE #126: Add a reinstall option
Hi,

On our site, when we want to reinstall a node, we run successively 3 commands : 
- `aii-shellfe --remove $node` to clean things up.
- `aii-shellfe --configure $node` to create all needed files, entries, ...
- `aii-shellfe --install $node` to set the node to install at next boot.

Therefore, I think it could be useful if we have a one-liner `aii-shellfe --reinstall $node` option to do the same.

Inspired by #124

stdweird:
Fixes #126


stdweird:
completely without unittests (but tried locally)

jrha:
LGTM

PR #138, opened 2015-08-19
aii-ks: spma uses boolean_yes_no for proxy enabling

stdweird:
When the `spma` `proxy` attribute is used to toggle proxy configuration, it's value is ignored (only existence is checked), leading to wrong configuration or in best case, some `uninitialized value` warnings.

### DIFF
--- a/aii-ks/src/test/resources/kickstart_noproxy.pan
+++ b/aii-ks/src/test/resources/kickstart_noproxy.pan
@@ -1,0 +1,11 @@
+object template kickstart_noproxy;
+
+include 'kickstart';
+
+prefix "/software/components/spma";
+# beware of crappy legacy schema
+"proxy" = "no"; 
+
+"proxyhost" = "proxy.server,LOCALHOST";  # comma-separated list of proxy hosts
+"proxyport" = "1234"; # yes, this is a string in the spma schema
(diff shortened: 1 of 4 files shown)


aii-ks/src/test/resources/kickstart_proxy.pan
```diff
@@ -0,0 +1,11 @@
+object template kickstart_proxy;
+
+include 'kickstart';
+
+prefix "/software/components/spma";
+# beware of crappy leagcy schema
```
jrha: **legacy

aii-ks/src/test/resources/kickstart_noproxy.pan
```diff
@@ -0,0 +1,11 @@
+object template kickstart_noproxy;
+
+include 'kickstart';
+
+prefix "/software/components/spma";
+# beware of crappy leagcy schema
```
jrha: **legacy

jrha:
I don't understand what this is for from the title, could you add a description?

stdweird:
i updated the description, mind that this is only slightly better behaviour then spma itself, which completely ignores the `proxy` attribute.

jrha:
Thanks!

PR #141, opened 2015-09-11
ks: fix unittest regexps due to new output from ncm-lib-blockdevices

stdweird:
Changes introduced in https://github.com/quattor/ncm-lib-blockdevices/pull/51 lead to new output, so the regexps in the unittests needed modification


stdweird:
@jrha found by build_all_repos

jrha:
Thanks

PR #143, opened 2015-09-30
aii-opennebula: include support to set different RPC cloud domains

### LINKED ISSUE #142: aii-opennebula: include support to set different cloud domains into /etc/aii/opennebula.conf
At the moment ONE AII uses `/etc/aii/opennebula.conf` to access to a single RPC endpoint:
```cfg
[rpc]
password=
user=
```
Include support to be able to use different cloud domains if it is required, as example:
```cfg
[rpc]
...
[DOMAIN]
...
```
If `[DOMAIN]` is not set AII should still use the `[rpc]` section to keep backwards compatibility.

alvarosimon:
* Fix #142 : include support to set different cloud domains into /etc/aii/opennebula.conf

### DIFF
--- a/aii-opennebula/src/main/perl/opennebula.pm
+++ b/aii-opennebula/src/main/perl/opennebula.pm
@@ -44,7 +44,7 @@ sub make_one
     $config = Config::Tiny->read($filename);
     if (exists($config->{$domainname})) {
         $rpc = $domainname;
-        $main::this_app->info ("Detected configfile RPC section: [$domainname]");
+        $main::this_app->info ("Detected configfile RPC section: [$rpc]");
     };
     my $port = $config->{$rpc}->{port} || 2633;
     my $host = $config->{$rpc}->{host} || "localhost";
--- a/aii-opennebula/src/main/perl/opennebula.pm
+++ b/aii-opennebula/src/main/perl/opennebula.pm
@@ -44,6 +44,7 @@ sub make_one
     $config = Config::Tiny->read($filename);
     if (exists($config->{$domainname})) {
         $rpc = $domainname;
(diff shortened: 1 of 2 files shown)


aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -38,12 +39,16 @@ sub make_one
...
+    my $domainname = $data->getElement (DOMAINNAME)->getValue;
 
     $config = Config::Tiny->read($filename);
-    my $port = $config->{rpc}->{port} || 2633;
-    my $host = $config->{rpc}->{host} || "localhost";
-    my $user = $config->{rpc}->{user} || "oneadmin";
-    my $password = $config->{rpc}->{password};
+    if (exists($config->{$domainname})) {
+        $rpc = $domainname;
+    };
```
stdweird: add a log message indictae which section was used

aii-opennebula/src/main/perl/opennebula.pod
```diff
@@ -116,6 +118,40 @@ Following package dependencies should be installed to run the AII:
...
 
+Set OpenNebula endpoints RPC connector /etc/aii/opennebula.conf
+
+It must include at least one RPC endpoint and password.
+
+By default ONE AII uses oneadmin user.
+
+It is also possible to set a different endpoint for each VM domain, as example:
+
+=over
```
stdweird: don't use a pod list here, just a raw text field. indent eacch line with 4 spaces and this should work

alvarosimon:
@stdweird fixed remarks. I have tested this in gastly using shuppet VMs.

jrha:
LGTM @apdibbo?

apdibbo:
LGTM

PR #144, opened 2015-10-05
aii-ks: when selinux is disabled in ks, don't install selinux policy rpms

stdweird:


stdweird:
@jnovy thanks for the suggestion

PR #152, opened 2015-12-04
kickstart: add excludepkgs/includepkgs support

### LINKED ISSUE #151: AII not using excludedPackages?
Hi,

Following this thread: https://github.com/quattor/ncm-ncd/issues/52
It looks like AII understands specifying exact versions of packages through `pkg_replace` but does not take `excludepkgs` directives from repository declarations into consideration.

One consequence is AII not able to run ccm-fetch when some repository overrides perl-common-sense and puts files in another place.

Since I'm able to reproduce this error on my side, I can run tests and give more information upon request.

Cheers,
Julien

stdweird:
Fixes #151

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -1307,6 +1307,13 @@ EOF
 priority=$repo->{priority}
 EOF
         }
+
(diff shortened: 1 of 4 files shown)


stdweird:
@jrha it's a small patch, so i tagged it for 15.12

PR #156, opened 2016-01-12
aii: add opennebula and freeipa rpm

kwaegema:


jrha:
Why isn't there `config.pan` for aii-freeipa?

PR #160, opened 2016-02-13
aii-ks: support yum.conf options

stdweird:
Related to https://github.com/quattor/configuration-modules-core/pull/710


jrha:
@stdweird this needs updating now that quattor/configuration-modules-core#710 is merged.

stdweird:
@jrha done

PR #161, opened 2016-02-14
ks: generate ccm.conf like ncm-ccm

stdweird:
Related https://github.com/quattor/configuration-modules-core/pull/709


stdweird:
It's odd we never noticied that the initial ccm.conf is so different from the actual one (esp wrt `json_typed`). anyway, long overdue...

aii-ks/src/main/perl/ks.pm
```diff
@@ -73,9 +68,16 @@ use constant { KS               => "/system/aii/osinstall/ks",
                INIT_SPMA_IGN_DEPS   => "/system/aii/osinstall/ks/init_spma_ignore_deps",
            };
 
+# These attributes are not configuration options
+# profile is handled separately
+use constant CCM_CONFIG_NOOPTIONS => qw(
```
ned21: This is brittle to the component schema changing and introducing new options.  There's also the possibility of breaking your ccm if you specify group_readable but the group isn't available during the install. I would prefer a whitelist of known config options, and default owner-only.

stdweird: `group_readable` should not be treated differently imho. if it fails here, it will also fail in `ncm-ccm` (and currently, an unknown group is non-fatal).
wrt the whitelist, this is the `ncm-ccm` code; i can make it explicit and factor out the generation of the `ccm.conf` from `ncm-ccm` and call it here.

jrha:
@stdweird rebase?

stdweird:
@ned21 any further remarks?

PR #164, opened 2016-02-19
aii-opennebula: var without default

kwaegema:
without this, template-library-core fails for https://github.com/quattor/template-library-core/pull/99


alvarosimon:
@kwaegema ups! I forgot to include it, this variable is required if you want to force to remove the resources within remove hook...

stdweird:
@kwaegema rebase please
LGTM

kwaegema:
@stdweird rebased

PR #169, opened 2016-04-04
aii-ks: sha512 instead of md5 for password hashes

victor-mendoza:
md5 is too fast, therefore much easier to brute-force.

This changes the default password hashing algorithm to sha512.

### DIFF
--- a/aii-ks/src/main/pan/quattor/aii/ks/config.pan
+++ b/aii-ks/src/main/pan/quattor/aii/ks/config.pan
@@ -295,9 +295,9 @@ variable  AII_OSINSTALL_BOOTPROTO ?= 'dhcp';
 
 #
 # Options for authentication
-# defaults to using shadow passwords and MD5 hashing
+# defaults to using shadow passwords and sha512 hashing
 #
-variable AII_OSINSTALL_OPTION_AUTH ?= list ("enableshadow", "enablemd5");
+variable AII_OSINSTALL_OPTION_AUTH ?= list ("enableshadow", "passalgo=sha512");
 "/system/aii/osinstall/ks/auth" ?= AII_OSINSTALL_OPTION_AUTH;
 #
 # Firewall
--- a/aii-ks/src/main/pan/quattor/aii/ks/schema.pan
+++ b/aii-ks/src/main/pan/quattor/aii/ks/schema.pan
@@ -57,7 +57,7 @@ type structure_ks_logging = {
 type structure_ks_ks_info = {
     "ackurl"	: type_absoluteURI
     "acklist"	? type_absoluteURI[]
-    "auth"		: string[] = list ("enableshadow", "enablemd5")
+    "auth" : string[] = list ("enableshadow", "passalgo=sha512")
     "bootloader_location" : string = "mbr"
     "bootloader_append" ? string
     "bootloader_password" ? string
--- a/aii-ks/src/main/perl/ks.pod
+++ b/aii-ks/src/main/perl/ks.pod
@@ -34,8 +34,8 @@ If acklist is set, then ackurl is not used.
 
 =item * auth : string[]
 
-List of options for authentication. Defaults to ("enableshadow", "enablemd5").
-i.e. enable shadow passwords and MD5 hashing.
+List of options for authentication. Defaults to ("enableshadow", "passalgo=sha512").
+i.e. enable shadow passwords and sha512 hashing.
 
 =item * bonding : boolean
 
@@ -458,8 +458,8 @@ the network via DHCP.
 
 =item * AII_OSINSTALL_OPTION_AUTH : string[]
 
-List of options for the authorization system. By default, MD5 shadowed
-passwords are enabled, this is C<list ("enableshadow", "enablemd5")>
+List of options for the authorization system. By default, sha512 shadowed
+passwords are enabled, this is C<list ("enableshadow", "passalgo=sha512")>
 
 =item * AII_OSINSTALL_OPTION_FIREWALL ? firewal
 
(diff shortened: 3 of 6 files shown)


aii-ks/src/main/perl/ks.pod
```diff
@@ -34,7 +34,7 @@ If acklist is set, then ackurl is not used.
 
 =item * auth : string[]
 
-List of options for authentication. Defaults to ("enableshadow", "enablemd5").
+List of options for authentication. Defaults to ("enableshadow", "passalgo=sha512").
 i.e. enable shadow passwords and MD5 hashing.
```
ned21: s/MD5/sha512/ in this line.

aii-ks/src/main/perl/ks.pod
```diff
@@ -459,7 +459,7 @@ the network via DHCP.
 =item * AII_OSINSTALL_OPTION_AUTH : string[]
 
 List of options for the authorization system. By default, MD5 shadowed
```
ned21: MD5 => sha512

aii-ks/src/main/pan/quattor/aii/ks/config.pan
```diff
@@ -297,7 +297,7 @@ variable  AII_OSINSTALL_BOOTPROTO ?= 'dhcp';
 # Options for authentication
 # defaults to using shadow passwords and MD5 hashing
 #
-variable AII_OSINSTALL_OPTION_AUTH ?= list ("enableshadow", "enablemd5");
+variable AII_OSINSTALL_OPTION_AUTH ?= list ("enableshadow", "passalgo=sha512");
```
ned21: ugh - not relevant to this PR but the fact this needs updating in multiple places is reason why the way AII uses global variables like this is bad practice. :( Wish we could eliminate it.

ned21:
Can you confirm this has been tested on SL5 as well as 6? In what version is the sha512 support introduced? We should potentially call that out in the release notes if it ends support for older versions.

victor-mendoza:
I've tested this on EL6 and EL7, sha512 support was added in EL5.2.
Managers of systems running previous releases would need to set:
`
variable AII_OSINSTALL_OPTION_AUTH ?= list("enableshadow", "enablemd5");
`

ned21:
As long as it works on EL5 then I don't think we need to tag this PR as "backwards incompatible". Given the second commit is just a fixup,can you squash it into the first please? (and rebase against master)

victor-mendoza:
done

jrha:
In which case we should probably also change the default password hash!

`$1$ZAOkBwVp$Cs5cO5cfaqzH5AdZ/jpjP/` is md5, sha-512 would start with `$6`

jrha:
`$6$umMRGl.0$.yKa1bdvMBCyVGLv.BH0dCI4mIIPv7Km0NwU90fOyiDY8DkebYOfVwJUQrtaYMSKVA16p1Ov7UBONSAX5bt8J/` should do the job.

stdweird:
@jrha hey, that's my pincode! (but seriously, a default hash?)

jrha:
Or get rid of the defaul, but we have one... so pick an option...

jrha:
Or am I actually crazy and we got rid of it already?

jrha:
Pfft, yup, I'm insane.
Is it worth at least updating the unit test to use the sha512 hash?

aii-ks/src/main/pan/quattor/aii/ks/schema.pan
```diff
@@ -57,7 +57,7 @@ type structure_ks_logging = {
 type structure_ks_ks_info = {
     "ackurl"	: type_absoluteURI
     "acklist"	? type_absoluteURI[]
-    "auth"		: string[] = list ("enableshadow", "enablemd5")
+    "auth"		: string[] = list ("enableshadow", "passalgo=sha512")
```
stdweird: fix the whitespace/tab after `"auth"`

stdweird: @jrha are you proposing to but a regex on the string to match either the begin of sha512 or md5? (have no clue if it s feasible)

aii-ks/src/main/pan/quattor/aii/ks/schema.pan
```diff
@@ -57,7 +57,7 @@ type structure_ks_logging = {
...
     "bootloader_location" : string = "mbr"
     "bootloader_append" ? string
     "bootloader_password" ? string
```
jrha: We should have a type to ensure this is a valid hash and not something mangled.

victor-mendoza:
You're suggesting changing "veryverysecret" in the unit tests with a sha512 hash?

jrha:
We can go further, the password hash should be three parts, seperated by dollar signs. They are:

1. hash function
   * Single digit, I propose allowing 1 (md5), 5 (sha-256) and 6 (sha-512).
2. salt
   * Alphanumeric (base64) string
3. hash
   * Alphanumeric (base64) string, length dependent on hash function

It should be possible to test all this, but perhaps not in this PR.

stdweird:
@jrha open an issue in template-library-core to support such type? then it can be used here with scattering new types around

aii-ks/src/main/pan/quattor/aii/ks/schema.pan
victor-mendoza: You wan't me to fix the whitespace/tab only on the "auth" line?

stdweird: yes, just this line is enough. if you do larger search and replace, make sure you do it all in one separate whitespace-only commit

jrha:
Lovely, thanks!

Issue #145, opened 2015-10-15
option to need explicitly confirmation before wiping a system with aii


kwaegema:
it would be useful to have a feature you can set on specific hosts (eg storage nodes) to need specific confirmation before wiping it. eg. aii-shellfe could ask you to add the buildid before doing something to such a host (except for booting)

PR #172, opened 2016-04-06
aii-shellfe : add protected feature and move to module

kwaegema:
Originally adding extra feature to be able to protect certain hosts , fixing #145 
I moved the core code to an AII::Shellfe module to be able to test this.
Only newly added sub is `check_protected`, and this is called in the `cmds` sub as:
```
if ($cmd =~ m/^($PROTECTED_COMMANDS)$/) {
                %h = $self->check_protected(%h);
                if (!(%h)) {
                    $self->error('No nodes left to process after checking for protected hosts');
                    return ;
                }
```


kwaegema:
This is documented in quattor/types/aii.pan , pr https://github.com/quattor/template-library-core/pull/110

kwaegema:
Since @stdweird had a lot of work on top of this, I will close this PR

PR #181, opened 2016-06-07
aii-opennebula: Include regex support within aii config file

### LINKED ISSUE #180: aii-opennebula: Support regular expressions in aii/opennebula.conf
Include support to use sections with regular expressions within aii opennebula.conf file

alvarosimon:
* Fix #180 : Support regular expressions in aii/opennebula.conf

Now is possible to include regular expressions within `opennebula.conf` sections as example:

> [myhost]
> host=...
> password=...
> pattern = node2\d+


kwaegema:
what would happen if you specify the regex with the domainname, eg. [node220*.shuppet.os] ? (is there a test for it?)

alvarosimon:
@kwaegema No, I didn't include fqdns to avoid any confusion just hostnames regex, but we can check fqdns as well

alvarosimon:
@kwaegema I have included fqdn regex instead hostnames so now is also valid to include these regex:

> [node220.+]
> [node1\d+.example.com]

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -40,9 +40,19 @@ sub make_one
...
+    my $fqdn = "${hostname}.${domainname}";
 
     $config = Config::Tiny->read($filename);
-    if (exists($config->{$domainname})) {
+    foreach my $section (sort keys %{$config}) {
+        $main::this_app->info ("Found RPC section: $section");
+        my $pattern = $config->{$section}->{pattern} || "";
+        if ($fqdn =~ /^$pattern$/ and $rpc eq 'rpc') {
+            $rpc = $section;
+            $main::this_app->info ("Match pattern in RPC section: [$rpc]");
```
stdweird: add a `last;` to stop processing any further entries

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -40,9 +40,19 @@ sub make_one
 
     my $config = Config::Tiny->new;
     my $domainname = $data->getElement (DOMAINNAME)->getValue;
+    my $hostname = $data->getElement (HOSTNAME)->getValue;
+    my $fqdn = "${hostname}.${domainname}";
 
     $config = Config::Tiny->read($filename);
-    if (exists($config->{$domainname})) {
+    foreach my $section (sort keys %{$config}) {
+        $main::this_app->info ("Found RPC section: $section");
```
stdweird: `verbose` or `debug(1,` is enough

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -40,9 +40,19 @@ sub make_one
...
+        my $pattern = $config->{$section}->{pattern} || "";
```
stdweird: remove the `|| ''` and use a `if ($pattern && $fqdn =~ ...`

stdweird:
LGTM

PR #186, opened 2016-07-05
refactor aii-dhcp

stdweird:
Depends on #195


ned21:
b31028a and 41ce9 have no explanation of what they intend to achieve.

103ae0 looks interesting but I'm not quite sure what this is for or how it fully works. Is there a corresponding schema change somewhere for /system/aii? Where is the best place to document this? What is the protected id mentioned in the command line help.

679bb10 explain why this is necessary. Could it break anything?

049216 makes changes not covered by the commit log (removal of the BEGIN block)

795761 squash into 679bb10?

71d490f - not reviewed, ran out of time.

stdweird:
@ned21 most of your remarks are related to #172. the main motivation to move things to a module is to be able to unittest the code
@kwaegema i'll let you address the remarks in #172? if not, i can also do them here, but then your PR should be closed (rewriting git history is not going to be nice)

kwaegema:
@ned21 fixed the commit messages in #172  . Documentation is added for this in https://github.com/quattor/template-library-core/pull/110 
@stdweird I will close that pr then, since it is not yet reviewed

stdweird:
@ned21 please review or merge the depending PRs #190 and #191 first. rewriting history in git is not that nice.

stdweird:
@ned21 last bit :smile: i'll even handle comments on the old code

ned21:
Most of the change in 9cd1419 is refactoring rather than a functional change which makes the functional change hard to spot. Is it possible to split this into two commits, one for the refactoring only and one for the functional change?

Everything else looks good.

stdweird:
@ned21 i split the commit

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -671,144 +679,112 @@ sub fetch_profiles
...
-    while (my ($node, $st) = each (%nl)) {
-        $self->debug (1, "Removing $node");
-        my $lock = $self->lock_node($node);
-        if (! $lock) {
-            $self->error("aii-shellfe: couldn't acquire lock on $node for remove");
-            exit(1);
-        }
+    my ($self, $node, $st) = @_;
+
+    $self->iter_plugins($st, REMOVEMETHOD);
```
ned21: This appears to introduce a second call to ```$self->iter_plugins($st, REMOVEMETHOD);``` - is that intentional? I see it gets removed again in the next commit.

stdweird: certainly not intentional. it's a result from the interactive rebase. splitting commits is not that trivial. i'll try to fix it

stdweird:
@ned21 i cleaned up the splitted commits

PR #190, opened 2016-07-31
aii-shellfe : add protected feature and move to module

### LINKED ISSUE #145: option to need explicitly confirmation before wiping a system with aii

stdweird:
Adds feature to be able to protect certain hosts
Moved aii-shellfe code to AII::Shellfe module for unittesting.

Fixes #145 #193

This PR replaces #172


aii-core/src/main/perl/aii-shellfe
```diff
@@ -491,6 +493,10 @@ sub app_options
          HELP    => 'Do not update OS installer (e.g. kickstart) configuration',
          DEFAULT => undef },
 
+       { NAME    => "$PROTECTED_OPTION=s",
+         HELP    => 'confirm id for reinstalling protected nodes',
```
ned21: Is there a more detailed explanation of how this works? i.e. how do I figure out what the correct id is? Is is the FQDN of the host or something else?

aii-core/src/main/perl/aii-shellfe
```diff
@@ -375,6 +375,8 @@ use constant DHCPPATH  => '/system/aii/dhcp';
 use constant MAC       => '--mac';
 
 
+Readonly my $PROTECTED_COMMANDS   => 'remove|configure|install';
+Readonly my $PROTECTED_OPTION   => 'protectid';
```
ned21: How about unprotectid to make it clearer that this option removes protection? Or --remove-protection-id ?

ned21:
GH regards my comments as outdated but the file was later refactored so the content is the same.

aii-core/src/main/perl/aii-shellfe
stdweird: the commandline option could be renamed to `--confirm`, as the protection feature basically requires you to add a `--yes-i-know-what-i-m-doing confirmationstring`.

aii-core/src/main/perl/aii-shellfe
stdweird: it's set in the profile, and printed in the error when you try to use certain options that require it.
we use the buildid, to avoid people to run something from history `ctrl-r` style.
we'll add some documentation after we settle on a name :smile:

stdweird:
@ned21 in case you are reviewing by commit, i renamed the protected commadnline in the refactor commit https://github.com/quattor/aii/pull/190/commits/e05442f7a7c3a29364192cb20dc05b19bd6991d6
the help now reads `confirm when removing/configuring/(re)installing protected nodes. A node is protected when /system/aii/protected is set, and its value is the confirmation to use`
there really not that much more to explain i think.

if you have remarks about the Shellfe.pm code which is now completely reviewable due to the code move, i'll adress those in #186, as that one also already contains some Shellfe.pm cleanup.
i'm aware this is a bit of a mess,  the alternative is that we only review #186, when that one is merged, also #190 and #191 will get merged.

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -0,0 +1,955 @@
...
+       { NAME    => 'nonbp',
+         HELP    => 'Do not update Network Boot Protocol (e.g. pxe) configuration',
+         DEFAULT => undef },
+
+       { NAME    => 'noosinstall',
+         HELP    => 'Do not update OS installer (e.g. kickstart) configuration',
+         DEFAULT => undef },
+
+       { NAME    => "$PROTECTED_OPTION=s",
+         HELP    => 'confirm when removing/configuring/(re)installing protected nodes. A node is protected when /system/aii/protected is set, and its value is the confirmation to use.',
```
ned21: How about: ```required when removing/configuring/(re)installing protected nodes otherwise the operation will fail. A node is protected when /system/aii/protected is set, and the value of that path must be passed to this parameter.```

Which makes me wonder: does this only accept a single argument or can multiple values of protected be passed e.g. as a list?

ned21:
Got it - thanks for the explanation I was utterly confused! One suggested improvement to the document string, plus a clarification required, then this is good to merge.

aii-core/src/main/perl/Shellfe.pm
stdweird: @ned21 i modified the help.
currently this is a single value. how would a list be used?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -0,0 +1,955 @@
...
+       { NAME    => 'nonbp',
+         HELP    => 'Do not update Network Boot Protocol (e.g. pxe) configuration',
+         DEFAULT => undef },
+
+       { NAME    => 'noosinstall',
+         HELP    => 'Do not update OS installer (e.g. kickstart) configuration',
+         DEFAULT => undef },
+
+       { NAME    => "$PROTECTED_OPTION=s",
+         HELP    => 'required when removing/configuring/(re)installing protected nodes otherwise the operation will fail. A node is protected when /system/aii/protected is set, and the value of that path must be passed to this parameter.',
```
ned21: GH swallowed your last comment about single v multiple values. Given only a single value is possible, may I suggest one further tweak?

required when removing/configuring/(re)installing a protected node otherwise the operation will fail. A node is protected when /system/aii/protected is set, and the value of that path must be passed to this parameter. Consequently only one protected node can be modified in a single shellfe command invocation.'

stdweird: ah, now i see. well, i'll rephrase it to `Consequently only those (protected) nodes with same confirmation can be modified in a single shellfe command invocation`
* you can actually modify more than one node
* even with confirmation given, you can still modify any not-protected node

ned21: Thanks, that seems very clear. I'll look out for the updated PR.

stdweird:
@ned21 done forgot to `push --force`....

PR #185, opened 2016-07-04
Shellfe: profileinfo from dir using dir:// cdburl

stdweird:
Allows to use `aii-shellfe` without (or with outdated) `profiles-info.xml`.
Depends on #186

PR #191, opened 2016-07-31
Shellfe: profileinfo from dir using dir:// cdburl

stdweird:
Allows to use aii-shellfe without (or with outdated) profiles-info.xml.
Depends on #190 
Replaces #185


aii-core/src/main/perl/aii-shellfe
```diff
@@ -491,6 +493,10 @@ sub app_options
          HELP    => 'Do not update OS installer (e.g. kickstart) configuration',
          DEFAULT => undef },
 
+       { NAME    => "$PROTECTED_OPTION=s",
+         HELP    => 'confirm id for reinstalling protected nodes',
```
ned21: Is there a more detailed explanation of how this works? i.e. how do I figure out what the correct id is? Is is the FQDN of the host or something else?

aii-core/src/main/perl/aii-shellfe
```diff
@@ -375,6 +375,8 @@ use constant DHCPPATH  => '/system/aii/dhcp';
 use constant MAC       => '--mac';
 
 
+Readonly my $PROTECTED_COMMANDS   => 'remove|configure|install';
+Readonly my $PROTECTED_OPTION   => 'protectid';
```
ned21: How about ```unprotectid``` to make it clearer that this option removes protection? Or ```--remove-protection-id``` ?

stdweird:
Replaced by #195 due to changes in git history in #190

PR #192, opened 2016-08-03
aii-opennebula: Include OpenNebula v5.0 support

### LINKED ISSUE #187: aii-opennebula: Include OpenNebula 5.x support
ONE 5.0 includes some new features and the VM template was changed. aii-opennebula must check the ONE version through the RPC endpoint and change the VM template accordingly.

alvarosimon:
* Fix #187 : Include OpenNebula 5.x support

### DIFF
--- a/aii-opennebula/src/main/perl/opennebula.pm
+++ b/aii-opennebula/src/main/perl/opennebula.pm
@@ -90,7 +90,7 @@ sub process_template
     
     my $tree = $config->getElement('/')->getTree();
     if ((defined $oneversion) and ($oneversion >= version->new("4.9.0"))) {
-        $tree->{system}->{opennebula}->{boot} = $oneversion;
+        $tree->{system}->{opennebula}->{boot} = "nic0,disk0";
         $main::this_app->info("BOOT section updated to support OpenNebula version >= 5.0.0");
     };
 
--- a/aii-opennebula/src/main/perl/opennebula.pm
+++ b/aii-opennebula/src/main/perl/opennebula.pm
@@ -24,6 +24,11 @@ use constant MAXITER => 20;
 use constant TIMEOUT => 30;
 use constant MINIMAL_ONE_VERSION => version->new("4.8.0");
 
+
+
+use constant BOOT_V4 => [qw(network hd)];
+use constant BOOT_V5 => [qw(nic0 disk0)];
+
 # a config file in .ini style with minmal 
 #   [rpc]
 #   password=secret
@@ -90,8 +95,10 @@ sub process_template
     
     my $tree = $config->getElement('/')->getTree();
     if ((defined $oneversion) and ($oneversion >= version->new("4.9.0"))) {
-        $tree->{system}->{opennebula}->{boot} = "nic0,disk0";
+        $tree->{system}->{opennebula}->{boot} = BOOT_V5;
         $main::this_app->info("BOOT section updated to support OpenNebula version >= 5.0.0");
+    } else {
+        $tree->{system}->{opennebula}->{boot} = BOOT_V4;
     };
 
     my $tpl = CAF::TextRender->new(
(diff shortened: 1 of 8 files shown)


stdweird:
@alvarosimon please rebase

alvarosimon:
@stdweird  PR sync with master. Now is ready to be reviewed

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -86,9 +86,14 @@ sub make_one
 # Detect and process ONE templates
 sub process_template
 {
-    my ($self, $config, $tt_name) = @_;
+    my ($self, $config, $tt_name, $oneversion) = @_;
     
     my $tree = $config->getElement('/')->getTree();
+    if ((defined $oneversion) and ($oneversion >= version->new("4.9.0"))) {
+        $tree->{system}->{opennebula}->{boot} = $oneversion;
```
stdweird: why don't you pass an arrayref here and do the join in the TT file?

alvarosimon: I was also thinking about this. Probably in the future more things will change with another update in a different section. Maybe we can create a new function called `update_config_tree` or something like that, this function could update, depending on the version detected by the RPC, the conf tree, so we will set within this method `$tree->{system}->{opennebula}->{$section}=$version` if it's required and return the modified tree. After these changes we only use the TT file to write a specific conf for a specific version, if the version is not set we write the default setup, instead to include all the logic in the TT file. 

What do you think? I'm not sure what is the best way to handle this...

stdweird: mayeb in the future yes, but for now, it's only the boot section, so i'd keep it here.

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -86,9 +91,16 @@ sub make_one
 # Detect and process ONE templates
 sub process_template
 {
-    my ($self, $config, $tt_name) = @_;
+    my ($self, $config, $tt_name, $oneversion) = @_;
     
     my $tree = $config->getElement('/')->getTree();
+    if ((defined $oneversion) and ($oneversion >= version->new("4.9.0"))) {
```
stdweird: change version to `5.0.0` as the message mentions

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -86,9 +91,16 @@ sub make_one
...
+        $tree->{system}->{opennebula}->{boot} = BOOT_V5;
+        $main::this_app->info("BOOT section updated to support OpenNebula version >= 5.0.0");
+    } else {
+        $tree->{system}->{opennebula}->{boot} = BOOT_V4;
```
stdweird: add message that this is boot section for older than 5.0.0

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -94,10 +94,11 @@ sub process_template
     my ($self, $config, $tt_name, $oneversion) = @_;
     
     my $tree = $config->getElement('/')->getTree();
-    if ((defined $oneversion) and ($oneversion >= version->new("4.9.0"))) {
+    if ((defined $oneversion) and ($oneversion >= version->new("5.0.0"))) {
         $tree->{system}->{opennebula}->{boot} = BOOT_V5;
-        $main::this_app->info("BOOT section updated to support OpenNebula version >= 5.0.0");
+        $main::this_app->info("BOOT section set to support OpenNebula versions >= 5.0.0");
```
stdweird: verbose is enough

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -94,10 +94,11 @@ sub process_template
...
     } else {
+        $main::this_app->info("BOOT section set to support OpenNebula versions < 5.0.0");
```
stdweird: verbose is enough

alvarosimon:
@stdweird remarks addressed

PR #195, opened 2016-08-07
Shellfe: profileinfo from dir using dir:// cdburl

stdweird:
Allows to use aii-shellfe without (or with outdated) `profiles-info.xml`.
Depends on #190
Replaces #191 which replaced #185


stdweird:
@ned21 this one is next towards #186

Issue #200, opened 2016-08-24
aii-core: issue when the main tree has more than one entry in "/system/aii/nbp"


Pansanel:
Hi,

We need to set for disabling bonding during boot time:

```
"/system/aii/nbp/ks/info/bonding" = false;
```

After compilation, the configuration tree contains a supplementary element `ks`:

```
     "nbp": {
        "ks": {
          "info": {
            "bonding": false
          }
        },
        "pxelinux": {
          "initrd": "sl680_x86_64/initrd.img",
          "kernel": "sl680_x86_64/vmlinuz",
          "ksdevice": "12:34:56:65:43:21",
          "kslocation": "http://quattor.example.org/ks/wn1.example.org.ks",
          "label": "Scientific Linux 680 (x86_64)"
        }
      },
```

In this case, the `aii-shellfe` fails with the following error:

```
[DEBUG] Instantiating ks
[DEBUG] no method Install available for plugin ks
```

This is caused by the line https://github.com/quattor/aii/blob/master/aii-core/src/main/perl/Shellfe.pm#L363 where the perl script is running the ks plugin instead of pxelinux plugin!
Setting the plugin name from a position in a key list is not safe. Is there a case where the name of the plugin is not **pxelinux**?

jrha:
This is what I would expect! Setting
```pan
"/system/aii/nbp/ks/info/bonding" = false;
```
to me, disables bonding during kickstart, not boot.

What behaviour are you actually trying to achieve?

jouvin:
@Pansanel I am not sure what you want to do but is not it `/system/aii/osinstall/ks` that you would like to set. I have the feeling that `/system/aii/nbp` can accept any key and that just `nbp` has a schema, thus the reason it compiled.

jouvin:
I just checked, `/system/aii` is basically defined as `extensible` in `quattor/types/aii.pan`. And then the parts really used are assigned a schema that restricts what is possible. The end result is that you cannot catch misdefined paths... Personally I never liked this and this is probably something that should be revised at some point... At least to restrict what is extensible.

Pansanel:
Using ```"/system/aii/nbp/ks/info/bonding" = false;``` makes ```aii-shellfe --install``` failing silently (you only see "no method Install available for plugin ks" with debug enabled.
If nbp can accept any key, we should modify:
```
my $modulename = (keys (%rm))[0];
```
with
```
my $modulename = 'pxelinux';
```
[...]

PR #202, opened 2016-08-26
ks: schema: bonding is optional and best-effort when not defined

stdweird:
Solves the reported error in https://github.com/quattor/aii/issues/200 when a macaddress is used as ksdevice


stdweird:
@Pansanel this will only handle the error. if something else was also going wrong when using a macaddress as ksdevice, please give some feedback as it should not occur. (you also do not have to not disable bonding anymore)

stdweird:
rebased

jrha:
LGTM

PR #205, opened 2016-08-31
aii-opennebula: configure fix

### LINKED ISSUE #199: AII-Opennebula: hook behaviour of --configure
Right now by default --configure removes and recreates the image if it exists. This should not be done with --configure but only with --remove.

### LINKED ISSUE #204: aii-opennebula: http rpc endpoint is hardcoded
RPC url http is hardcoded from aii-opennebula code and it is not able to use https endpoints:

>  url      => "http://$host:$port/RPC2",

change aii configuration to include an url endpoint, as example:

> [rpc]
> password=...
> url=https://hostname:2634

### LINKED ISSUE #208: aii-opennebula: add some error msgs to opennebula_vmtemplate_vnet type checks
something like this 

```
--- a/cfg/standard/quattor/aii/opennebula/schema.pan
+++ b/cfg/standard/quattor/aii/opennebula/schema.pan
@@ -55,6 +55,7 @@ type opennebula_vmtemplate_vnet = string{} with {
     # check if all entries in the map have a network interface
     foreach (k;v;SELF) {
         if (! exists("/system/network/interfaces/"+k)) {
+            error(format("entry %s in the vnet map has no network interface", k)) ;
             return(false);
         };
     };
@@ -65,6 +66,7 @@ type opennebula_vmtemplate_vnet = string{} with {
              (! match('^(Bridge|OVSBridge)$', v['type'])) # these special types are OK 
              )
             ) {
+            error(format("interface %s has no entry in the vnet map", k)) ;
             return(false);
         };
     };
```

alvarosimon:
* Fix #199 : hook behaviour of --configure 
 * Fix #204 : http rpc endpoint is hardcoded
 * Fix #208 : add some error msgs to opennebula_vmtemplate_vnet type checks

`--configure` does not remove any image or template, only `--remove` does

### DIFF
--- a/aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
+++ b/aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
@@ -55,6 +55,7 @@ type opennebula_vmtemplate_vnet = string{} with {
     # check if all entries in the map have a network interface
     foreach (k;v;SELF) {
         if (! exists("/system/network/interfaces/"+k)) {
+            error(format("entry: %s in the vnet map is not available from /system/network/interfaces tree", k));
             return(false);
         };
     };
@@ -65,6 +66,7 @@ type opennebula_vmtemplate_vnet = string{} with {
              (! match('^(Bridge|OVSBridge)$', v['type'])) # these special types are OK 
              )
             ) {
+            error(format("/system/network/interfaces/%s has no entry in the vnet map", k));
             return(false);
         };
     };
@@ -75,12 +77,14 @@ type opennebula_vmtemplate_datastore = string{} with {
     # check is all entries in the map have a hardrive
     foreach (k;v;SELF) {
         if (! exists("/hardware/harddisks/"+k)) {
+            error(format("/hardware/harddisks/%s has no entry in the datastores map", k));
             return(false);
         };
     };
     # check if all interfaces have an entry in the map
     foreach (k;v;value("/hardware/harddisks")) {
         if (! exists(SELF[k])) {
+            error(format("entry: %s in the datastore map is not available from /hardware/harddisks tree", k));
             return(false);
         };
     };
--- a/aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
+++ b/aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
@@ -56,7 +56,6 @@ type opennebula_vmtemplate_vnet = string{} with {
     foreach (k;v;SELF) {
         if (! exists("/system/network/interfaces/"+k)) {
             error(format("entry: %s in the vnet map is not available from /system/network/interfaces tree", k));
-            return(false);
         };
     };
     # check if all interfaces have an entry in the map
--- a/aii-opennebula/src/main/perl/opennebula.pm
+++ b/aii-opennebula/src/main/perl/opennebula.pm
@@ -29,10 +29,10 @@ use constant MINIMAL_ONE_VERSION => version->new("4.8.0");
 use constant BOOT_V4 => [qw(network hd)];
 use constant BOOT_V5 => [qw(nic0 disk0)];
 
-# a config file in .ini style with minmal 
+# a config file in .ini style with minmal
 #   [rpc]
 #   password=secret
-sub make_one 
+sub make_one
 {
     my ($self, $data) = @_;
     my $filename = AII_OPENNEBULA_CONFIG;
@@ -62,18 +62,22 @@ sub make_one
         $rpc = $domainname;
         $main::this_app->info ("Detected configfile RPC section: [$rpc]");
     };
-    my $port = $config->{$rpc}->{port} || 2633;
-    my $host = $config->{$rpc}->{host} || "localhost";
+    my $port = $config->{$rpc}->{port};
+    my $host = $config->{$rpc}->{host}; #Deprecated
+    my $url = $config->{$rpc}->{url} || "http://localhost:2633";
     my $user = $config->{$rpc}->{user} || "oneadmin";
     my $password = $config->{$rpc}->{password};
+
+    $url = "http://$host:2633" if ($host);
+
 
     if (! $password ) {
         $main::this_app->error("No password set in configfile $filename. Section [$rpc]");
         return;
     }
-    
+
     my $one = Net::OpenNebula->new(
-        url      => "http://$host:$port/RPC2",
+        url      => "$url/RPC2",
         user     => $user,
         password => $password,
         log => $main::this_app,
@@ -92,7 +96,7 @@ sub make_one
 sub process_template
 {
     my ($self, $config, $tt_name, $oneversion) = @_;
-    
+
     my $tree = $config->getElement('/')->getTree();
     if ((defined $oneversion) and ($oneversion >= version->new("5.0.0"))) {
         $tree->{system}->{opennebula}->{boot} = BOOT_V5;
@@ -142,8 +146,8 @@ sub get_fqdn
 }
 
 # It gets the image template from tt file
-# and gathers image names format: <fqdn>_<vdx> 
-# and datastore names to store the new images 
+# and gathers image names format: <fqdn>_<vdx>
+# and datastore names to store the new images
 sub get_images
 {
     my ($self, $config) = @_;
@@ -280,12 +284,12 @@ sub remove_and_create_vm_images
     # Check created/removed image lists
     if ($remove) {
         my $diff = $self->check_vm_images_list(\@rimages, \@qimages);
-        if ($diff) { 
+        if ($diff) {
             $main::this_app->error("Removing these VM images: ", join(', ', @qimages));
         }
     } else {
         my $diff = $self->check_vm_images_list(\@nimages, \@qimages);
-        if ($diff) { 
+        if ($diff) {
             $main::this_app->error("Creating these VM images: ", join(', ', @qimages));
         }
     }
@@ -386,7 +390,7 @@ sub remove_and_create_vn_ars
             my $arinfo = $t->get_ar(%ar_opts);
             if ($remove) {
                 $self->remove_vn_ars($one, $arinfo, $vnet, $ardata, $t);
-            } else { 
+            } else {
                 $self->create_vn_ars($one, $vnet, $ardata, $t);
             }
        }
@@ -472,7 +476,7 @@ sub stop_and_remove_one_vms
 sub remove_and_create_vm_template
 {
     my ($self, $one, $fqdn, $createvmtemplate, $vmtemplate, $permissions, $remove) = @_;
-    
+
     # Check if the vm template already exists
     my @existtmpls = $one->get_templates(qr{^$fqdn$});
 
@@ -496,7 +500,7 @@ sub remove_and_create_vm_template
             $main::this_app->info("No QUATTOR flag found for VM template: ",$t->name);
         }
     }
-    
+
     if ($createvmtemplate && !$remove) {
         my $templ = $one->create_template($vmtemplate);
         $main::this_app->debug(1, "New ONE VM template name: ",$templ->name);
@@ -604,7 +608,7 @@ sub install
     my $onhold = $tree->{onhold};
     $main::this_app->info("Start VM onhold flag is set to: $onhold");
     my $permissions = $self->get_permissions($config);
-        
+
     my $fqdn = $self->get_fqdn($config);
 
     # Set one endpoint RPC connector
@@ -661,7 +665,7 @@ sub install
 }
 
 # Performs Quattor post_reboot
-# ACPID service is mandatory for ONE VMs 
+# ACPID service is mandatory for ONE VMs
 sub post_reboot
 {
     my ($self, $config, $path) = @_;
--- a/aii-opennebula/src/main/perl/opennebula.pm
+++ b/aii-opennebula/src/main/perl/opennebula.pm
@@ -15,21 +15,19 @@ use Template;
 use Config::Tiny;
 use Net::OpenNebula 0.307.0;
 use Data::Dumper;
+use Readonly;
 
-use constant TEMPLATEPATH => "/usr/share/templates/quattor";
-use constant AII_OPENNEBULA_CONFIG => "/etc/aii/opennebula.conf";
-use constant HOSTNAME => "/system/network/hostname";
-use constant DOMAINNAME => "/system/network/domainname";
-use constant MAXITER => 20;
-use constant TIMEOUT => 30;
-use constant MINIMAL_ONE_VERSION => version->new("4.8.0");
-use constant ONE_DEFAULT_URL => 'http://localhost:2633/RPC2';
-use constant ONE_DEFAULT_PORT => 2633;
-use constant ONE_DEFAULT_USER => "oneadmin";
-
-
-use constant BOOT_V4 => [qw(network hd)];
-use constant BOOT_V5 => [qw(nic0 disk0)];
+Readonly my $AII_OPENNEBULA_CONFIG => "/etc/aii/opennebula.conf";
+Readonly my $HOSTNAME => "/system/network/hostname";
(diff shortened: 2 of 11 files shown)


alvarosimon:
@kwaegema this PR solves the remove issue, please take a look. It was tested in altaria.

ned21:
@alvarosimon - d41027e seems to do more than just more than removing hardcoded rpc. Can you please amend the commit log to explain all the changes? (or split the commit into separate ones for refactoring/functional changes if the other changes are just clean up). Same for bea31da: it appears to touch more than just the logging calls.

It's best practice too if a unit test fix up (acb88b6) is squashed into the commit that broke the build so that we don't have individual commits that are not atomic (e.g. if we need to revert 1 commit, it's not obvious that another commit may also need to be reverted). 

The commit log message for a8acfac is so brief that it does not make much sense to me but maybe that's because I am not familiar with opennebula so I'll let @kwaegema comment on whether a more expansive documentation of the change is required.

alvarosimon:
@ned21 you are right,  I did some extra changes and the commit logs are not clear enough for the reviewer, sorry for that, I will update the PR and log commits asap

alvarosimon:
@ned21 I think that now is a bit more clear, sorry again.
This reminds me of something... :smile: http://xkcd.com/1296/

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -62,18 +62,22 @@ sub make_one
...
     };
-    my $port = $config->{$rpc}->{port} || 2633;
-    my $host = $config->{$rpc}->{host} || "localhost";
+    my $port = $config->{$rpc}->{port};
+    my $host = $config->{$rpc}->{host};
+    my $url = $config->{$rpc}->{url} || "http://localhost:2633";
     my $user = $config->{$rpc}->{user} || "oneadmin";
     my $password = $config->{$rpc}->{password};
 
+    $url = "http://$host:2633" if ($host);
```
kwaegema: this overwrites the url set by `$config->{$rpc}->{url}` if only $host is set?
One should probably specify `url NAND (host OR port)`  since you have defaults, otherwise this can be ambiguous, (and I would throw an error in that case..) Then, if the `url` is set, take that one, if not keep old behavior to keep backwards compatibility ?

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -62,18 +62,22 @@ sub make_one
...
+
     if (! $password ) {
         $main::this_app->error("No password set in configfile $filename. Section [$rpc]");
         return;
     }
-    
+
     my $one = Net::OpenNebula->new(
-        url      => "http://$host:$port/RPC2",
+        url      => "$url/RPC2",
```
kwaegema: you still have `host` and `port` variables, but port is `never` used

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -29,10 +29,10 @@ use constant MINIMAL_ONE_VERSION => version->new("4.8.0");
 use constant BOOT_V4 => [qw(network hd)];
 use constant BOOT_V5 => [qw(nic0 disk0)];
 
-# a config file in .ini style with minmal 
+# a config file in .ini style with minmal
```
kwaegema: minimal? minimal what? :smile:

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -294,36 +301,35 @@ sub remove_and_create_vm_images
...
-    if (!$remove) {
-        if ($self->is_one_resource_available($one, "image", $imagename)) {
-            $main::this_app->error("Image: $imagename is already used and AII hook: image is not set. ",
-                                    "Please remove this image first.");
-        } else {
+    if ($self->is_one_resource_available($one, "image", $imagename)) {
+        $main::this_app->warn("Image: $imagename is already available from OpenNebula. ",
+                              "Please remove this image first if you want to generate a new one from scratch.");
+        return;
+    } else {
```
kwaegema: not needed since you do a `return`

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -472,20 +479,20 @@ sub stop_and_remove_one_vms
...
-    
+
     # Check if the vm template already exists
     my @existtmpls = $one->get_templates(qr{^$fqdn$});
 
     foreach my $t (@existtmpls) {
         if ($t->{extended_data}->{TEMPLATE}->[0]->{QUATTOR}->[0]) {
-            if ($createvmtemplate) {
+            if ($remove) {
                 # force to remove and create the template again
```
kwaegema: this only removes it, no?

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -496,7 +503,7 @@ sub remove_and_create_vm_template
             $main::this_app->info("No QUATTOR flag found for VM template: ",$t->name);
         }
     }
-    
+
     if ($createvmtemplate && !$remove) {
```
kwaegema: I don't understand how this works, but probably  I don't completely get what the `$remove` flag means

kwaegema: OK, I already get it :smile:

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -472,20 +479,20 @@ sub stop_and_remove_one_vms
 sub remove_and_create_vm_template
```
kwaegema: The name of the sub is confusing right now. It is more a remove_or_create_vm_template, depending on the $remove flag. I would make it more clear that what this sub does (removing or creating) is depending on the context it is called (configure or remove)

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -267,25 +271,28 @@ sub get_resource_instance
 
 sub remove_and_create_vm_images
```
kwaegema: Same remakr about sub name and some doc context  maybe

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -705,20 +708,20 @@ sub remove
...
 
+    my $vmtemplatetxt = $self->get_vmtemplate($config, $oneversion);
+    if ($vmtemplatetxt) {
+        $self->remove_and_create_vm_template($one, $fqdn, 1, $vmtemplatetxt, undef, $rmvmtemplate);
+    }
+
     my %images = $self->get_images($config);
     if (%images) {
-        $self->remove_and_create_vm_images($one, 1, \%images, undef, $rmimage);
+        $self->remove_and_create_vm_images($one, undef, \%images, undef, $rmimage);
```
kwaegema: Is there a reason this is called when `$rmimage` is false?

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -705,20 +708,20 @@ sub remove
         $self->stop_and_remove_one_vms($one,$fqdn);
     }
 
+    my $vmtemplatetxt = $self->get_vmtemplate($config, $oneversion);
+    if ($vmtemplatetxt) {
+        $self->remove_and_create_vm_template($one, $fqdn, 1, $vmtemplatetxt, undef, $rmvmtemplate);
```
kwaegema: Should this be called when  $rmvmtemplate is false. The current behavior now is to update the template then when it is false (which I guess is not needed in a remove context)

kwaegema:
@alvarosimon I added some remarks.

Maybe it is also nice to describe somewhere (in the pod?) how the opennebula parts (vm, image, template,..) map to the typical aii parts `remove` , `configure`, `install` . 
Something like 
remove: stop vm, remove vm/image/template/.. 
configure: creating / updating .. which resources (again) ? 
install: only instantiate machine created by resources of configure ?

Unless that is already described somewhere I don't know of :smile:

aii-opennebula/src/main/perl/opennebula.pm
alvarosimon: @kwaegema ah ok I will change this, in any case the ambiguity disappears due to the opennebula metaconfig aii setup that only allows urls setup: https://github.com/quattor/configuration-modules-core/pull/788

alvarosimon:
@kwaegema thanks for the review! I have addressed your remarks, but if something is missing or it is not clear please let me know :smile:

aii-opennebula/src/main/perl/opennebula.pod
```diff
@@ -96,8 +99,6 @@ All these paths are relative to /system/aii/hooks/remove
 
 Stop running VM
 
-=over 4
```
wdpypere: if you remove the ```=over``` here, you'll have to remove one of the  ```=back``` a bit further as well.

wdpypere: hmm, don't we validate pods in AII tests?

```code```
[wdpypere@scyther ~]$ podchecker opennebula.pod 
*** ERROR: =back without =over at line 112 in file opennebula.pod
opennebula.pod has 1 pod syntax error.
```

alvarosimon: Ups.. thanks I forgot that, now the syntax is ok

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -269,7 +271,7 @@ sub get_resource_instance
     return;
 }
 
-sub remove_and_create_vm_images
+sub remove_or_create_vm_images
```
kwaegema: add docstring

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -476,7 +478,7 @@ sub stop_and_remove_one_vms
 
 # Creates and removes VM templates
 # $createvmtemplate hook forces to remove/create
-sub remove_and_create_vm_template
+sub remove_or_create_vm_template
```
kwaegema: update docstring

aii-opennebula/src/main/perl/opennebula.pod
kwaegema: It is included with the components in configuration-modules-core, but not here it seems. @alvarosimon you can add the file pod-syntax.t to the test dir with contents 

```
use Test::More;
use Test::Pod;

my @dirs = qw(target/lib/perl target/doc/pod);
all_pod_files_ok(all_pod_files(@dirs));
```

alvarosimon: Ah yes.. pod-syntax.t is missing... :disappointed: I will include that thanks!

alvarosimon:
@wdpypere @kwaegema remarks addressed

kwaegema:
I'm still thinking the $url override by a $host is weird, but if metaconfig enforces this (and everyone should use metaconfig :smile: ), I guess it doesn't matter. besides that, LGTM!

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -63,17 +65,21 @@ sub make_one
         $main::this_app->info ("Detected configfile RPC section: [$rpc]");
     };
     my $port = $config->{$rpc}->{port} || 2633;
-    my $host = $config->{$rpc}->{host} || "localhost";
+    my $host = $config->{$rpc}->{host};
+    my $url = $config->{$rpc}->{url} || "http://localhost:2633/RPC2";
```
stdweird: the defaults (inlc port and user) should be readonlys, like `Readonly my ONE_DEFAULT_URL => 'http://localhost:2633/RPC2';`

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -63,17 +65,21 @@ sub make_one
...
     my $user = $config->{$rpc}->{user} || "oneadmin";
     my $password = $config->{$rpc}->{password};
 
+    # Keep backwards compatibility
+    $url = "http://$host:$port/RPC2" if ($host);
```
stdweird: add a log warning here that the url is redefined

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -63,17 +65,21 @@ sub make_one
...
+
     if (! $password ) {
         $main::this_app->error("No password set in configfile $filename. Section [$rpc]");
         return;
     }
-    
+
     my $one = Net::OpenNebula->new(
-        url      => "http://$host:$port/RPC2",
+        url      => "$url",
```
stdweird: no quotes around `$url`?

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -462,30 +472,32 @@ sub stop_and_remove_one_vms
             $main::this_app->info("Running VM will be removed: ",$t->name);
             $t->delete();
         } else {
-            $main::this_app->info("No QUATTOR flag found for Running VM: ",$t->name);
+            $main::this_app->info("No QUATTOR flag found for Running VM: ", $t->name);
```
stdweird: please add in the message that you stop processing and other `@runingvms` here?
but is the return really necessary?

aii-opennebula/src/main/perl/opennebula.pod
stdweird: see also #206

aii-opennebula/src/main/perl/opennebula.pm
alvarosimon: No  `return` is not necessary in this case, I will remove that

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -23,6 +23,7 @@ use constant DOMAINNAME => "/system/network/domainname";
 use constant MAXITER => 20;
 use constant TIMEOUT => 30;
 use constant MINIMAL_ONE_VERSION => version->new("4.8.0");
+use constant ONE_DEFAULT_URL => 'http://localhost:2633/RPC2';
```
kwaegema:
```
use Readonly;
Readonly my ONE_DEFAULT_URL => 'http://localhost:2633/RPC2';
```
like https://github.com/quattor/configuration-modules-core/blob/master/ncm-ceph/src/main/perl/Ceph/crushmap.pm#L42

alvarosimon: @stdweird should we switch to `Readonly` the rest of the constant values as well? `MINIMAL_ONE_VERSION`, `TIMEOUT`...

stdweird: up to you, but we should not mix constant and Readonly, that's just confusing

alvarosimon:
@stdweird remarks addressed

aii-opennebula/src/main/perl/opennebula.pm
```diff
@@ -347,7 +362,7 @@ sub is_timeout
     my ($self, $one, $resource, $name) = @_;
     eval {
         local $SIG{ALRM} = sub { die "alarm\n" };
-        alarm TIMEOUT;
+        alarm $TIMEOUT;
```
piojo: Doesn't `$SIG{ALRM}` cause a `die`?

alvarosimon: @piojo from http://perldoc.perl.org/functions/alarm.html it seems that you should use `eval/die` to be sure that the timeout will work (but I'm not a perl guru, I'm not sure why..)

aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
```diff
@@ -55,6 +55,7 @@ type opennebula_vmtemplate_vnet = string{} with {
     # check if all entries in the map have a network interface
     foreach (k;v;SELF) {
         if (! exists("/system/network/interfaces/"+k)) {
+            error(format("entry: %s in the vnet map is not available from /system/network/interfaces tree", k));
```
alvarosimon: @piojo ah ok, I have changed that, thanks for the tip

jrha:
LGTM. @piojo are you happy with this now?

PR #207, opened 2016-09-15
Shellfe: Correct handling when fetchProfile returns an error

stdweird:


aii-core/src/main/perl/Shellfe.pm
```diff
@@ -657,7 +657,7 @@ sub fetch_profiles
             next;
         }
         my @err = $fh->fetchProfile();
-        if ($err[0] != SUCCESS) {
+        if (! $err[0] || $err[0] == -1) {
```
ned21: -1 looks like a magic number, can it be replaced by a constant? Or better yet, let's do ```|| $err[0] < 0``` because then in future release we can replace the undef return with -2 and make all errors negative numbers and all successful results positive ones.

cf. [fetchProfile doc](https://github.com/quattor/CCM/blob/master/src/main/perl/Fetch.pm#L125)

stdweird: i sincerely hope we never add `-2`, and that we even get rid of `-1`. failures should be easy to test, the reason why something when wrong can be handled via e.g. a fail `attribute` (even better, retrun undef in case of network issue, and `die()` for everything else).

anyway, i'll use the `$ERROR` readonly that is actually returned (altough the pod says `-1`).

stdweird: hmm, rereading fetchProfile, it actually never returns more than one element...

stdweird: @ned21 i cleaned it up a bit more.

stdweird: @ned21 https://github.com/quattor/CCM/pull/129

ned21:
LG, although please can you modify the commit message to reflect the actual change? ("Correct handling when fetchProfile returns an error" or similar)

I also raised a question in quattor/CCM#129 about how undef gets returned.

stdweird:
@ned21 remark fixed

PR #211, opened 2016-09-30
aii-pxelinux: substitute LOCALHOST on profile append

ajf8:
We do this for the kickstart URL but it can be useful for parameters
specified in templates using "append" from the profile.

### DIFF
--- a/aii-pxelinux/src/main/perl/pxelinux.pm
+++ b/aii-pxelinux/src/main/perl/pxelinux.pm
@@ -332,9 +332,13 @@ sub pxe_ks_append
             push(@append,"nameserver=$ns");
         }
     }
-        
-    push(@append, $t->{append}) if $t->{append};
 
+    my $custom_append = $t->{append};
+    if ($custom_append) {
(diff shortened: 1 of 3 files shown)


ajf8:
Any idea why this might be on the builder?

not ok 1 - use NCM::Component::pxelinux;
Stacktrace

(proc): exec(/usr/sbin/selinuxenabled): No such file or directory
[VERB] Log options --verbose

stdweird:
@ajf8 hmm, why don't you set the corect value in the profile directly?
if you want to keep this "feature", we need documnetation and unittests to make sure noone removes this in the future.

ned21:
I assume this is because we have up to 4 boot servers running AII so we need a template to allow the current boot server to insert the correct (current) hostname?

stdweird:
@ned21 after readong the code, it's already used before (alothough alos not documented/tested?) i was confused with `LOCALHOST`, should have been `AIISERVER` to make it clearer.

stdweird:
@ajf8 thx!

PR #213, opened 2016-10-05
Remove stratuslab-srv01.lal.in2p3.fr:8081 as a distribution (maven) repository

ajf8:
It is no longer updated, maven central is used instead. The fact this repository is on a non-standard port can hang maven on some networks.


jrha:
LGTM

PR #218, opened 2016-10-31
aii-ks: fix logging of pre and post script actions for EL7

### LINKED ISSUE #120: aii-ks 15.4: missing CR in messages displayed at the console
I don't know if this is specific to our HW but with `aii-ks` 15.4 (may be also the case with previous version...) when installing EL7, with graphic mode disabled (the default), all the messages printed during KS pre/post scripts are missing a CR (only the LF is taken into account). The result is hardly readable. Something like:

```
message text first line
                       message text second line
                                                yet another line
```

Is it a know problem? Are we the only one with it?

jouvin:
Fixes #120

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -692,8 +692,14 @@ sub log_action {
     }
 
     if ($consolelogging) {
+        # In EL7, /dev/console doesn't add the carriage return for each line feed (new line)
+        # resulting in unreadable messages on the console. /dev/pts/0 must be used instead.
+        # But on prior versions, /dev/pts/0 doesn't exist at installation time and cannot be used.
+        # The following code allows to use /dev/pts/0 if it exists else to revert to /dev/console.
+        push(@logactions, "console='/dev/console'");
+        push(@logactions, "[ -c /dev/pts/0 ] && console='/dev/pts/0'");
         push(@logactions, '# Make sure messages show up on the serial console',
-                          "tail -f $logfile > /dev/console &");
+                          "tail -f $logfile > \$console &");
     }
 
     push(@logactions,"drainsleep=$drainsleep"); # add trailing newline
--- a/aii-ks/src/test/perl/kickstart_logging.t
+++ b/aii-ks/src/test/perl/kickstart_logging.t
@@ -33,7 +33,7 @@ like($fh, qr{^%packages\s--ignoremissing\s--resolvedeps\n^package\n^package2\n^b
 # logaction tests
 my $logaction = NCM::Component::ks::log_action($cfg, 'mylogfile', 1);
 like($logaction, qr{^exec\s>mylogfile\s2>&1}, 'start with exec redirection'); # no multiline search!
-like($logaction, qr{^tail\s-f\smylogfile\s>\s/dev/console\s&}m, 'console logging enabled');
+like($logaction, qr{^tail\s-f\smylogfile\s>\s\$console\s&}m, 'logfile displayed on console');
 
 like($logaction, qr{^wait_for_network\slogserver}m, 'Insert sleep to make sure network is up');
 like($logaction, qr{^\(tail\s-f\smylogfile.*?usleep.*?\snc\s-u\slogserver\s514\)\s&$}m, 'netcat udp logsending');
--- a/aii-ks/src/test/perl/kickstart_nologging_just_method_and_protocol.t
+++ b/aii-ks/src/test/perl/kickstart_nologging_just_method_and_protocol.t
@@ -33,8 +33,8 @@ my $logaction = NCM::Component::ks::log_action($cfg, 'mylogfile', 1);
 like($logaction, qr{^exec\s>mylogfile\s2>&1}, 'start with exec redirection'); # no multiline search!
 
 # hmm, first test might be too specific in case of subtle changes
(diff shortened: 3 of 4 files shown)


aii-ks/src/main/perl/ks.pm
```diff
@@ -692,13 +692,12 @@ sub log_action {
     }
 
     if ($consolelogging) {
-        push(@logactions, '# Make sure messages show up on the serial console',
-                          "tail -f $logfile > /dev/console &");
+        push(@logactions, "tee $logfile", "tee /dev/pts/0");
```
stdweird: i guess `tail -f $logfile > /dev/pts/0` doesn't do the trick?

aii-ks/src/main/perl/ks.pm
```diff
@@ -692,13 +692,12 @@ sub log_action {
...
     }
 
-    push(@logactions,"drainsleep=$drainsleep"); # add trailing newline
-    push(@logactions,''); # add trailing newline
-    return join("\n", @logactions)
+#    push(@logactions,"drainsleep=$drainsleep"); # add trailing newline
+#    push(@logactions,''); # add trailing newline
+    return join(" | ", @logactions)
```
stdweird: the remote logging feature generates after exec
```
(tail -f /tmp/pre-log.log | awk '{print "<190>AII: "$0; fflush(); system("usleep 1000 >& /dev/null");}' | nc -u gastly.gastly.os 515) &
```
i'm not sure how to replace this more pipes

aii-ks/src/main/perl/ks.pm
jouvin: Good suggestion! I didn't try, I will do it. Clearly if at the end this is just the matter of changin `console` to `pts/0`, it would be an easy change...

jouvin:
@stdweird you're right! The problem is that `/dev/console` on EL7 has a new behaviour and that `/dev/pts/0` must be used instead... Unfortunately, `/dev/pts/0` doesn't exist at this point in the installation on EL6... so it is not as easy as replacing one by the other. The last commit provides a fixes tested successfully on EL6 (SL) and EL7 (Cent0S).

jouvin:
One of the unit test checking for the present of `/dev/console` still has to be fixed!

aii-ks/src/main/perl/ks.pm
```diff
@@ -692,8 +692,14 @@ sub log_action {
     }
 
     if ($consolelogging) {
+        # In EL7, /dev/console doesn't add the carriage return for each line feed (new line)
+        # resulting in unreadable messages on the console. /dev/pts/0 must be used instead.
+        # But on prior versions, /dev/pts/0 doesn't exist at installation time and cannot be used.
+        # The following code allows to use /dev/pts/0 if it exists else to revert to /dev/console.
+        push(@logactions, "console='/dev/console'");
```
stdweird: there is no harm in always setting these variables early on, even if no consolelogging is used. this will allow a cleaner if/else construct, because this is a bit ugly.

jouvin: I am not sure to understand your point. Why it would be clearer to have it earlier? This variable definition in the %pre script is useful only if you use console logging, there is  no real point in adding them otherwise. I would find it more confusing to have this defined unconditionally, both in the ks.pm code and the resulting %pre script...

jouvin:
I don't really understand why the unit test `kickstart_pre` is failing (`regexps/pre_noblock_logging`). Any attempt to do the same test in a test perl script seems to work... Help would be greatly appreciated!

BTW, all these unit tests seem very fragile... On my SL6 machine, many (16) subtests of `kickstart_pre` are failing as `kickstart_mounts`, despite they success on Jenkins. I have the feeling we rely on something very installation-specific...

jouvin:
I managed to get the unit test 'kickstart_pre` passing successfully after removing the following line from `regexps/pre_noblock_logging`:
```
^\[ -e /dev/pts/0 \] && console='/dev/pts/0'$
```

I am not sure what was wrong in it but I am not completely convinced of the value of testing each line this way...

aii-ks/src/main/perl/ks.pm
stdweird: it's a bit hackish. but i saw that i did similar hack for the wait time, so i'm not going to complain any further :smile:

stdweird:
the lines are tested to make sure they are not deleted/modified unintentionally. the test fails because you have `-c` instead of `-e`?
also, always run the tests with clean, and run them with verbose output, so we can help debugging.

stdweird:
inparticular, use https://github.com/quattor/quattor.github.com/issues/146#issuecomment-144681288 to run the tests.

jouvin:
@stdweird I must really apologize for this `-c` versus `-e` mistake... I changed the initial `-e` to `-c` but tested the regexp on a machine where I forgot to update `ks.pm`... Unit tests are great 😄

jouvin:
For the record, the problems with `kickstart_pre` unit tests not related to this PR have been understood: it was caused by a too old `ncm-lib-blockdevice`. See https://github.com/quattor/aii/pull/218#issuecomment-257830478.

stdweird:
@jouvin if you have tested this, can you remove the `[WIP]` in the title?

stdweird:
LGTM
@jouvin can you cleanup the description again? should probably only mention the issue it fixes

PR #237, opened 2016-12-07
aii-core: switch to CAF::Download::LWP

stdweird:
Requires https://github.com/quattor/CAF/pull/203


stdweird:
retest this please

PR #238, opened 2016-12-08
aii-opennebula: Add VM image driver option

### LINKED ISSUE #235: aii-opennebula: Add qcow2 image support
We want to use qcow2 images with opennebula.
@alvarosimon is looking to add this feature.

alvarosimon:
By default ONE uses raw format but it is possible to use qcow2 as well.
Notice that qcow2 is not supported by Ceph storage backends.

 * Fix #235 : Add qcow2 image support

### DIFF
--- a/aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
+++ b/aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
@@ -188,6 +188,8 @@ type opennebula_vmtemplate = {
     "ignoremac" ? opennebula_ignoremac
     "graphics"  : string = 'VNC' with match (SELF, '^(VNC|SDL|SPICE)$')
     "diskcache" ? string with match(SELF, '^(default|none|writethrough|writeback|directsync|unsafe)$')
+    @{specific image mapping driver. qcow2 is not supported by Ceph storage backends}
+    "diskdriver" ? string with match(SELF, '^(raw|qcow2)$')
     "permissions" ? opennebula_permissions
     "pci" ? opennebula_vmtemplate_pci[]
     @{labels is a list of strings to group the VMs under a given name and filter them 
--- a/aii-opennebula/src/main/resources/imagetemplate.tt
+++ b/aii-opennebula/src/main/resources/imagetemplate.tt
@@ -11,6 +11,10 @@ DESCRIPTION = "QUATTOR image for [% fqdn %]: [% pair.key %]"
 [%- IF system.opennebula.labels.defined %]
 LABELS = "[% system.opennebula.labels.join(',') %]"
 [%- END %]
+[%- IF system.opennebula.diskdriver.defined %]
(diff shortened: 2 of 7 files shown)


alvarosimon:
@samary could you please try this in your testbed? I'm not able to generate qcow2 images because we are using Ceph in all our datastores.

samary:
It seems to require the FSTYPE set to qcow2 in the image template as well

samary:
It works for me ! Huge Thanks !

alvarosimon:
@samary thanks! 👍 
Included `FSTYPE`  variable that is required in ONE 4.12, this value is included automatically in ONE 5.x

aii-opennebula/src/main/pan/quattor/aii/opennebula/schema.pan
```diff
@@ -188,6 +188,8 @@ type opennebula_vmtemplate = {
     "ignoremac" ? opennebula_ignoremac
     "graphics"  : string = 'VNC' with match (SELF, '^(VNC|SDL|SPICE)$')
     "diskcache" ? string with match(SELF, '^(default|none|writethrough|writeback|directsync|unsafe)$')
+    @{specific image mapping driver. qcow2 is not supported by Ceph storage backends}
```
jrha: If qcow2 is not supported by Ceph, can you validate that it has not been used?

alvarosimon: Mmm.. that will require some extra work. At this moment to use the opennebula-aii we are mapping the VM hard drives to specific datastores, as example:

```
"datastore" = dict(
     "vda", "my_datastore1",
     "vdb", "my_datastore2");
```
so the aii will try to generate each image in each datastore, but if the datastore does not exist it will show an error message, at this moment does not detect if the datastore exits or what kind of datastore are you requesting (NFS,Ceph..), the aii just tries to use the datastore that are you asking for... so if you want to generate a qcow2 image and your datastore is located in a Ceph storage backend the aii will show an error message in the same way and the image is not generated and the VM is not started.
What we can do is to include a check from in the opennebula-aii logic to detect first of all  if the datastore is there and then what kind of datastore are you asking for to display a nicer error message that you are trying to use the wrong storage type.. I can open a ticket to include this.

jrha: If it's very hard then it's probably not worth it.

alvarosimon: @jrha this feature is not very hard to add so I have included this in our TODO list: 
https://github.com/quattor/configuration-modules-core/issues/1014

stdweird:
@alvarosimon is this code also in the component PR? can you otherwsie close or move any opennebula issue to the component repo? this is getting confusing

alvarosimon:
@stdweird yes this fix is already in  the new component PR

PR #225, opened 2016-11-04
Add UEFI support through Grub-based PXE loader

jouvin:
Fixes #216.

Fixes #258.

Also contributes to #255.

PR #239, opened 2016-12-12
AII:Shellfe: Add option to run methods in parallel

### LINKED ISSUE #241: EL7 AII with mutiple nodes stops dhcpd due to too many restarts
```
Job for dhcpd.service failed because start of the service was attempted too often. See "systemctl status dhcpd.service" and "journalctl -xe" for details.
To force a start use "systemctl reset-failed dhcpd.service" followed by "systemctl start dhcpd.service" again.
```

aii should be a lot smarter wrt restarts

kwaegema:
Enabling parallelism reduces number of daemon restarts and hence reduces time taken to apply changes.

- Adds an option to run part of the code in parallel: the plugins except for dhcp. 
- DHCP changes are now gathered using the AII::DHCP module instead of calling aii-dhcp. 
- DHCP is now started twice at most (in remove and configure phase) . This is done BEFORE the remove and configure (earlier this was done after other plugins for remove cmd)
- Unused options `restartcmd` and `dhcpconf` were removed. A new option `dhcpcfg` is provided, which makes it possible to overwrite the config file for AII::DHCP module
- Requires Perl > 5.10.1 — effectively removing support for EL5.
- Fixes #241

### DIFF
--- a/aii-core/src/main/perl/DHCP.pm
+++ b/aii-core/src/main/perl/DHCP.pm
@@ -422,7 +422,7 @@ sub new_configure_entry
                 MORE_OPT   => $add_txt,
              } );
         $self->debug(2, "AII::DHCP: add new entry: host = $host fqdn = $fqdn mac = $mac");
-        $self->debug(3, "AII::DHCP: add new entry: additional opts: $add_txt");
+        $self->debug(3, "AII::DHCP: add new entry: host = $host additional opts: $add_txt");
     }
 
     return(0);
@@ -461,8 +461,7 @@ sub read_input
         # get input data
         my $fh = CAF::FileReader->new($filename, log => $self);
         if ($EC->error()) {
-            $self->error("AII::DHCP: configurelist error: " .
-                             "file access error $filename");
+            $self->error("AII::DHCP: configurelist error: file access error $filename");
             $error +=1;
         } else {
             $self->debug(2, "AII::DHCP: reading nodes to configure from file: $filename");
@@ -511,10 +510,10 @@ sub update_and_restart
     # restart dhcpd daemon
     if ($changed) {
         if ($self->option('norestart')) {
-            $self->verbose("AII::DHCP: config changed but norestart set!");
+            $self->verbose("AII::DHCP: config $filename changed but norestart set!");
         } else {
             my $cmd = $self->option('restartcmd');
-            $self->debug(1, "AII::DHCP: restarting dhcpd daemon using cmd '$cmd'");
+            $self->info(1, "AII::DHCP: config $filename changed, restarting dhcpd daemon using cmd '$cmd'");
             # expects an arayref
             $self->restart_daemon([split(/\s+/, $cmd)]);
         } 
@@ -524,8 +523,9 @@ sub update_and_restart
     return 1;
 }
 
-# return true if dhcp config need changes
-sub nodes_to_change {
+# return true if nodes need to be configured or removed
+sub nodes_to_change 
+{
     my $self = shift;
     return  return (@{$self->{NTC}} + @{$self->{NTR}}) ? 1 : 0;
 }
@@ -544,9 +544,9 @@ sub configure
 
     # update dhcpd configuration file
     if ($self->nodes_to_change()) {
-		return 1 if (!$self->update_and_restart()) ;
+        return 1 if (!$self->update_and_restart()) ;
     } else {
-        $self->debug(1, "aii-nbp: there are no changes to dhcpd configuration to make");
+        $self->debug(1, "AII:DHCP: no nodes to configure, no changes to dhcp configuration");
     }
 
     return 0;
--- a/aii-core/src/main/perl/DHCP.pm
+++ b/aii-core/src/main/perl/DHCP.pm
@@ -131,14 +131,14 @@ sub restart_daemon
 {
     my ($self, $cmd) = @_;
 
-    $self->debug(3, "aii-dhcp: restarting daemon dhcpd");
+    $self->debug(3, "AII::DHCP: restarting daemon dhcpd");
 
     my $output = CAF::Process->new($cmd, log => $self)->output();
     if ($?) {
-        $self->error("aii-dhcp: error restarting dhcp daemon: $output");
+        $self->error("AII::DHCP: error restarting dhcp daemon: $output");
         return(1);
     } else {
-        $self->verbose("aii-dhcp: daemon restarted: $output");
+        $self->verbose("AII::DHCP: daemon restarted: $output");
     }
 
     return(0);
@@ -176,7 +176,7 @@ sub update_dhcp_config_file
     my @netandmasks = ($text =~ /\n\s*subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)/g);
 
     if ($#netandmasks % 2 == 0) {
-        $self->error("aii-dhcp: syntax error on dhcpd config file: " .
+        $self->error("AII::DHCP: syntax error on dhcpd config file: " .
                      "netmask/network missing in subnet declaration");
         return(1, '');
     }
@@ -189,7 +189,7 @@ sub update_dhcp_config_file
                 ST_NET  => $netandmasks[$i],
                 ST_MASK => $netandmasks[$i+1],
               });
-        $self->verbose("aii-dhcp: found subnet $netandmasks[$i] " .
+        $self->verbose("AII::DHCP: found subnet $netandmasks[$i] " .
                        "mask $netandmasks[$i+1]");
     }
 
@@ -237,10 +237,10 @@ sub update_dhcp_config_file
 
                 $newnodes .= "$indent}\n";
                 if ( $subnet_defined ) {
-                    $self->verbose("aii-dhcp: added node $node->{NAME} ".
+                    $self->verbose("AII::DHCP: added node $node->{NAME} ".
                                        "to subnet $net->{ST_NET}");
                 } else {
-                    $self->verbose("aii-dhcp: added node $node->{NAME} (no subnet specified)");
+                    $self->verbose("AII::DHCP: added node $node->{NAME} (no subnet specified)");
                 }
             }
 
@@ -248,7 +248,7 @@ sub update_dhcp_config_file
 
         # Insert the nodes in the current subnet
         if ($newnodes ne '') {
-            $self->debug(1, "aii-dhcp: newnodes=|$newnodes|\n");
+            $self->debug(1, "AII::DHCP: newnodes=|$newnodes|\n");
             if ( $subnet_defined ) {
                 $text =~ s/( \s*  subnet  \s+ \Q$net->{ST_NET}\E  \s+
                          netmask \s+ \Q$net->{ST_MASK}\E \s+ \{
@@ -326,14 +326,14 @@ sub new_remove_entry
 
     # Check host
     if (!defined($host) || $host eq '') {
-        $self->warn('aii-dhcp: missing hostname');
+        $self->warn('AII::DHCP: missing hostname');
         return(1);
     }
 
     my ($fqdn, @all_address) = (gethostbyname($host))[0,4];
     if (! @all_address) {
         # The array is empty => invalid name
-        $self->warn("aii-dhcp: invalid hostname to remove ($host), DNS lookup failed");
+        $self->warn("AII::DHCP: invalid hostname to remove ($host), DNS lookup failed");
         return(1);
     }
 
@@ -345,7 +345,7 @@ sub new_remove_entry
             IP   => $ip,
          } );
 
-    $self->debug(2, "aii-dhcp: mark $host (fqdn: $fqdn, ip: $ip) to remove");
+    $self->debug(2, "AII::DHCP: mark $host (fqdn: $fqdn, ip: $ip) to remove");
 
     return(0);
 }
@@ -360,24 +360,24 @@ sub new_configure_entry
 
     # Check hostname
     if (!defined($host) || $host eq '') {
-        $self->warn('aii-dhcp: missing hostname');
+        $self->warn('AII::DHCP: missing hostname');
         return(1);
     }
 
     my ($fqdn, @all_address) = (gethostbyname($host))[0,4];
     if ($#all_address < 0) {        # The array is empty => invalid name
-        $self->warn("aii-dhcp: invalid hostname to add ($host)");
+        $self->warn("AII::DHCP: invalid hostname to add ($host)");
         return(1);
     }
 
     # Check MAC address
     if (!defined($mac)) {
-        $self->warn("aii-dhcp: missing MAC address for host $host");
+        $self->warn("AII::DHCP: missing MAC address for host $host");
         return(1);
     }
 
     if ($mac !~ /^([[:xdigit:]]{2}[\:\-]){5}[[:xdigit:]]{2}$/) {
-        $self->warn("aii-dhcp: MAC address $mac not valid for host $host");
+        $self->warn("AII::DHCP: MAC address $mac not valid for host $host");
         return(1);
     }
 
@@ -392,7 +392,7 @@ sub new_configure_entry
             $tftpserver = Socket::inet_ntoa($all_tftp_address[0]);
         } else {
             # The array is empty => invalid name
-            $self->warn("aii-dhcp: invalid TFTP server ($tftpserver) for $host: DNS lookup failed");
+            $self->warn("AII::DHCP: invalid TFTP server ($tftpserver) for $host: DNS lookup failed");
             return(1);
         }
     }
@@ -403,7 +403,7 @@ sub new_configure_entry
     # check if the host entry already exists
     foreach my $item (@{$self->{NTC}}) {
         if($item->{FQDN} eq $fqdn) {
-            $self->debug(2, "aii-dhcp: new MAC entry for existing host = $host fqdn = $fqdn mac = $mac");
+            $self->debug(2, "AII::DHCP: new MAC entry for existing host = $host fqdn = $fqdn mac = $mac");
             $item->{MAC} .= " $mac";
             $mac = ""; # flag meaning object found
             last;
@@ -421,8 +421,8 @@ sub new_configure_entry
                 ST_IP_TFTP => $tftpserver,
                 MORE_OPT   => $add_txt,
              } );
-        $self->debug(2, "aii-dhcp: add new entry: host = $host fqdn = $fqdn mac = $mac");
-        $self->debug(3, "aii-dhcp: add new entry: additional opts: $add_txt");
+        $self->debug(2, "AII::DHCP: add new entry: host = $host fqdn = $fqdn mac = $mac");
+        $self->debug(3, "AII::DHCP: add new entry: additional opts: $add_txt");
     }
 
     return(0);
@@ -461,11 +461,11 @@ sub read_input
         # get input data
         my $fh = CAF::FileReader->new($filename, log => $self);
         if ($EC->error()) {
-            $self->error("aii-dhcp: configurelist error: " .
+            $self->error("AII::DHCP: configurelist error: " .
                              "file access error $filename");
             $error +=1;
         } else {
-            $self->debug(2, "aii-dhcp: reading nodes to configure from file: $filename");
+            $self->debug(2, "AII::DHCP: reading nodes to configure from file: $filename");
             foreach my $item (split(m/\n/, "$fh")) {
                 if ($item =~ m/^\s*([^#].*?)\s*$/) {
                     $error += $self->new_configure_entry(split(m/\s+/, $1));
@@ -480,10 +480,10 @@ sub read_input
 
         my $fh = CAF::FileReader->new($filename, log => $self);
         if ($EC->error()) {
-            $self->error("aii-dhcp: removelist error: file access error $filename");
+            $self->error("AII::DHCP: removelist error: file access error $filename");
             $error +=1;
         } else {
-            $self->debug(2, "aii-dhcp: reading nodes to remove from file: $filename");
+            $self->debug(2, "AII::DHCP: reading nodes to remove from file: $filename");
             foreach my $item (split(m/\n/, "$fh")) {
                 if ($item =~ m/^\s*([^#].*?)\s*$/) {
                     $error += $self->new_remove_entry($1);
@@ -501,25 +501,25 @@ sub update_and_restart
     my $self = shift;
 
     my $filename = $self->option('dhcpconf');
-    $self->debug(1, "aii-dhcp: going to update dhcpd configuration $filename");
+    $self->debug(1, "AII::DHCP: going to update dhcpd configuration $filename if needed");
     my ($error, $changed) = $self->update_dhcp_config($filename);
     if ($error) {
-        $self->error("aii-dhcp: failed to update dhcpd configuration $filename");
+        $self->error("AII::DHCP: failed to update dhcpd configuration $filename");
         return;
     }
 
     # restart dhcpd daemon
-    if ($self->option('norestart')) {
-        $self->verbose("aii-dhcp: dhcpd daemon do not restart (norestart set)");
-    } else {
-        if ($changed) {
+    if ($changed) {
+        if ($self->option('norestart')) {
+            $self->verbose("AII::DHCP: config changed but norestart set!");
+        } else {
             my $cmd = $self->option('restartcmd');
-            $self->debug(1, "aii-dhcp: restarting dhcpd daemon using cmd '$cmd'");
+            $self->debug(1, "AII::DHCP: restarting dhcpd daemon using cmd '$cmd'");
             # expects an arayref
             $self->restart_daemon([split(/\s+/, $cmd)]);
-        } else {
-            $self->verbose("aii-dhcp: no changes to $filename: daemon not restarted");
-        }
+        } 
+    } else {
+        $self->verbose("AII::DHCP: no changes to $filename: daemon not restarted");
     }
     return 1;
 }
@@ -527,7 +527,7 @@ sub update_and_restart
 # return true if dhcp config need changes
 sub nodes_to_change {
     my $self = shift;
-    return  (scalar(@{$self->{NTC}}) > 0) || (scalar(@{$self->{NTR}}) > 0);
+    return  return (@{$self->{NTC}} + @{$self->{NTR}}) ? 1 : 0;
 }
 
 # return 1 on failure, 0 on success
@@ -536,14 +536,14 @@ sub configure
     my $self = shift;
 
     # process command line options
-    $self->debug(1, "aii-dhcp: reading cmd line or input files");
+    $self->debug(1, "AII::DHCP: reading cmd line or input files");
     if ($self->read_input()) {
-        $self->error("aii-dhcp: failed to process cmd line or input files");
+        $self->error("AII::DHCP: failed to process cmd line or input files");
         return(1);
     }
 
     # update dhcpd configuration file
-    if($self->nodes_to_change() ) {
+    if ($self->nodes_to_change()) {
 		return 1 if (!$self->update_and_restart()) ;
     } else {
         $self->debug(1, "aii-nbp: there are no changes to dhcpd configuration to make");
--- a/aii-core/src/main/perl/DHCP.pm
+++ b/aii-core/src/main/perl/DHCP.pm
@@ -281,6 +281,7 @@ sub update_dhcp_config
 {
     my ($self, $filename) = @_;
 
+
     # Lock and load the current dhcp configuration file
     my $lockfile = $filename . ".lock";
     my $lock = CAF::Lock->new ($lockfile, log => $self);
@@ -495,6 +496,38 @@ sub read_input
     return $error;
 }
 
+# update the dhcp config file and restart daemon
+sub update_and_restart {
+	my $self = @_;
+
+    my $filename = $self->option('dhcpconf');
+    $self->debug(1, "aii-dhcp: going to update dhcpd configuration $filename");
+    my ($error, $changed) = $self->update_dhcp_config();
+    if ($error) {
+        $self->error("aii-dhcp: failed to update dhcpd configuration $filename");
+        return(1);
+    }
+
+    # restart dhcpd daemon
+    if ($self->option('norestart')) {
+        $self->verbose("aii-dhcp: dhcpd daemon do not restart (norestart set)");
+    } else {
+        if ($changed) {
+            my $cmd = $self->option('restartcmd');
+            $self->debug(1, "aii-dhcp: restarting dhcpd daemon using cmd '$cmd'");
+            # expects an arayref
+            $self->restart_daemon([split(/\s+/, $cmd)]);
+        } else {
+            $self->verbose("aii-dhcp: no changes to $filename: daemon not restarted");
+        }
+    }
+}
+
+# return true if dhcp config need changes
+sub nodes_to_change {
+    my $self = @_;
+    return  (scalar(@{$self->{NTC}}) > 0) || (scalar(@{$self->{NTR}}) > 0);
+}
 
 # return 1 on failure, 0 on success
 sub configure
@@ -509,29 +542,8 @@ sub configure
     }
 
     # update dhcpd configuration file
-    if( scalar(@{$self->{NTC}}) > 0 || scalar(@{$self->{NTR}}) > 0 ) {
-        my $filename = $self->option('dhcpconf');
-
-        $self->debug(1, "aii-dhcp: going to update dhcpd configuration $filename");
-        my ($error, $changed) = $self->update_dhcp_config($filename);
-        if ($error) {
-            $self->error("aii-dhcp: failed to update dhcpd configuration $filename");
-            return(1);
-        }
-
-        # restart dhcpd daemon
-        if ($self->option('norestart')) {
-            $self->verbose("aii-dhcp: dhcpd daemon do not restart (norestart set)");
-        } else {
-            if ($changed) {
-                my $cmd = $self->option('restartcmd');
-                $self->debug(1, "aii-dhcp: restarting dhcpd daemon using cmd '$cmd'");
-                # expects an arayref
-                $self->restart_daemon([split(/\s+/, $cmd)]);
-            } else {
-                $self->verbose("aii-dhcp: no changes to $filename: daemon not restarted");
-            }
-        }
+    if($self->nodes_to_change() ) {
+		$self->update_and_restart();
     } else {
         $self->debug(1, "aii-nbp: there are no changes to dhcpd configuration to make");
     }
--- a/aii-core/src/main/perl/DHCP.pm
+++ b/aii-core/src/main/perl/DHCP.pm
@@ -116,7 +116,6 @@ sub _initialize {
     unless ($self->SUPER::_initialize(@_)) {
         return;
     }
-
     # start using log file (could be done later on instead)
     $self->set_report_logfile($self->{'LOG'});
 
@@ -498,11 +497,11 @@ sub read_input
 
 # update the dhcp config file and restart daemon
 sub update_and_restart {
-	my $self = @_;
+	my $self = shift;
 
     my $filename = $self->option('dhcpconf');
     $self->debug(1, "aii-dhcp: going to update dhcpd configuration $filename");
-    my ($error, $changed) = $self->update_dhcp_config();
+    my ($error, $changed) = $self->update_dhcp_config($filename);
     if ($error) {
         $self->error("aii-dhcp: failed to update dhcpd configuration $filename");
         return(1);
@@ -525,7 +524,7 @@ sub update_and_restart {
 
 # return true if dhcp config need changes
 sub nodes_to_change {
-    my $self = @_;
+    my $self = shift;
     return  (scalar(@{$self->{NTC}}) > 0) || (scalar(@{$self->{NTR}}) > 0);
 }
 
--- a/aii-core/src/main/perl/DHCP.pm
+++ b/aii-core/src/main/perl/DHCP.pm
@@ -496,15 +496,16 @@ sub read_input
 }
 
 # update the dhcp config file and restart daemon
-sub update_and_restart {
-	my $self = shift;
+sub update_and_restart 
+{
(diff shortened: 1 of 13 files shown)


stdweird:
retest this please

kwaegema:
(Also tried with threads instead of forkmanager, but didn't get it to work https://github.com/kwaegema/aii/tree/threads)

stdweird:
swuash the commit and change the title of the PR to `parallel` instead of `threads`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -697,13 +701,38 @@ sub fetch_profiles
...
+                # retrieve data structure from child 
+                if (defined($data_struct_ref)) {
+                    $responses->{$id} = $data_struct_ref;
+                    $self->debug(5, "Running $cmd on $id had output"); 
+                }
+            }
+        );
+        return $pm;
+    } else {
+        return 0;
```
stdweird: just `return;`, unless the `0` really has a meaning

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -697,13 +701,38 @@ sub fetch_profiles
...
+    }
+}
+    
 # Wrapper to execute the commands in sorted manner
 no strict 'refs';
 foreach my $cmd (COMMANDS) {
     *{$cmd} = sub {
         my ($self, %node_states) = @_;
         my $method = "_$cmd";
+        my %responses = ();
```
stdweird: the `= ()` is not required

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -713,8 +742,23 @@ foreach my $cmd (COMMANDS) {
...
+            $self->debug(5, "Going to start $cmd on node $node");
+            if ($pm) { # start parallel execution in child
+                $pm->start($node) and next;
+            }
+
+            my $ec = $self->$method($node, $node_states{$node}) || 0;
+            my $res = { ec => $ec, method => $method, node => $node, mode => $pm ? 1 : 0 };
+
+            if ($pm) {
+                $pm->finish($ec, $res ); # Terminates the child process
```
stdweird: no space after res

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -713,8 +742,23 @@ foreach my $cmd (COMMANDS) {
...
                 };
             };
-            $self->$method($node, $node_states{$node});
+            $self->debug(5, "Going to start $cmd on node $node");
+            if ($pm) { # start parallel execution in child
+                $pm->start($node) and next;
+            }
+
+            my $ec = $self->$method($node, $node_states{$node}) || 0;
+            my $res = { ec => $ec, method => $method, node => $node, mode => $pm ? 1 : 0 };
```
stdweird: each key/value on one line

stdweird:
2 major remarks: 
* the reporting: it might be possibly mixed output (eg 2 child rpocs writng to the output file at the same time), but there's also no way to identify which output line comnes from which child; so we will have to enable the `--pid` log option wheneever parallel is enabled
* dhcpd will cause troubles. there's no easy way out of that one, dhcpd is not really suited for this. we could generate the leases files or swicth to kea, but restarting dhcpd upon config change is not ok. it's not a new option, but the parallel makes it really bad
i would propose to restructure the flow as follows: if dhcpd changes are enabled/exepected, do foreach command:
* (possibly in parallel): gather the relevant dhcp data from the profiles (like name, mac, ip , options)
 * make sure the CCM cache is not deleted, it is used again to do the non-dhcp stuff
* serial: loop over all options, modify dhcpd.conf and restart (once) if something changed
* (possibly in parallel): continue with non-dhcp stuff

this reduces the maximum number of dhcpd restarts to maximum 1 per command, something also non-parallel runs will benefit from.

kwaegema:
DHCP solution WIP. Tests included but not done yet

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
     return $error;
 }
 
+# update the dhcp config file and restart daemon
+sub update_and_restart {
```
stdweird: `{` on new line

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+	my $self = shift;
```
stdweird: fix indent

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+
+    my $filename = $self->option('dhcpconf');
+    $self->debug(1, "aii-dhcp: going to update dhcpd configuration $filename");
```
stdweird: the `aii-dhcp` is not correct. i would also not report `update` since you don't know yet if there will be a change or not.

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+    my ($error, $changed) = $self->update_dhcp_config($filename);
+    if ($error) {
+        $self->error("aii-dhcp: failed to update dhcpd configuration $filename");
```
stdweird: no `aii-shcp`

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+        return(1);
```
stdweird: document in the pod or comment that `1` means failure

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+    }
+
+    # restart dhcpd daemon
+    if ($self->option('norestart')) {
+        $self->verbose("aii-dhcp: dhcpd daemon do not restart (norestart set)");
+    } else {
+        if ($changed) {
+            my $cmd = $self->option('restartcmd');
+            $self->debug(1, "aii-dhcp: restarting dhcpd daemon using cmd '$cmd'");
```
stdweird: no `aii-dhcp`

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+    $self->debug(1, "aii-dhcp: going to update dhcpd configuration $filename");
+    my ($error, $changed) = $self->update_dhcp_config($filename);
+    if ($error) {
+        $self->error("aii-dhcp: failed to update dhcpd configuration $filename");
+        return(1);
+    }
+
+    # restart dhcpd daemon
+    if ($self->option('norestart')) {
+        $self->verbose("aii-dhcp: dhcpd daemon do not restart (norestart set)");
```
stdweird: report if anything changed or not.
i'd actually be better if the flow was
```
if changed:
  if restrta:
    restart
    versboe(changed and restarted)
  else:
     versboe(changed but norestart)
else:
  verbose(nothing to do)
```

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+            # expects an arayref
+            $self->restart_daemon([split(/\s+/, $cmd)]);
+        } else {
+            $self->verbose("aii-dhcp: no changes to $filename: daemon not restarted");
+        }
+    }
+}
+
+# return true if dhcp config need changes
+sub nodes_to_change {
```
stdweird: `{` on new line

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,6 +495,38 @@ sub read_input
...
+    my $self = shift;
+    return  (scalar(@{$self->{NTC}}) > 0) || (scalar(@{$self->{NTR}}) > 0);
```
stdweird: `return (@{$self->{NTC}} + @{$self->{NTR}}) ? 1 : 0`

aii-core/src/main/perl/DHCP.pm
```diff
@@ -509,29 +541,8 @@ sub configure
...
-            if ($changed) {
-                my $cmd = $self->option('restartcmd');
-                $self->debug(1, "aii-dhcp: restarting dhcpd daemon using cmd '$cmd'");
-                # expects an arayref
-                $self->restart_daemon([split(/\s+/, $cmd)]);
-            } else {
-                $self->verbose("aii-dhcp: no changes to $filename: daemon not restarted");
-            }
-        }
+    if($self->nodes_to_change() ) {
```
stdweird: space after `if`, no space before `)`

aii-core/src/main/perl/DHCP.pm
```diff
@@ -509,29 +541,8 @@ sub configure
...
+		$self->update_and_restart();
```
stdweird: indentation?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -30,8 +30,9 @@ use File::Path qw(mkpath rmtree);
 use File::Basename qw(basename);
 use DB_File;
 use Readonly;
-
+use Parallel::ForkManager 0.7.6;
```
stdweird: out of curiosity, what is so special about `0.7.6`. el6 has `0.7.9` in epel.
@ned21 @jrha @jouvin do we need el5 support for aii?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -464,16 +435,22 @@ sub dhcp
...
-       $self->error("Error when configuring $node: $output");
-    } else {
-        $self->debug(4, "Output: $output");
-    };
-    $self->debug(4, "End output");
+    my $ec;
+    if ("$cmd" eq CONFIGURE) {
+        my $opts = $st->{configuration}->getElement (DHCPOPTION)->getTree;
+        $self->debug (4, "Going to add dhcp entry of $node to configure");
+        $ec = $mgr->new_configure_entry($node, $mac, $opts->{tftpserver} // ';', $opts->{addoptions} // ());
```
stdweird: the `//` magic is not avail in el5, but it is in el6. so at least require perl 5.10.0.

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -464,16 +435,22 @@ sub dhcp
...
+    } elsif ("$cmd" eq REMOVEMETHOD) {
+        if ($st->{reinstall}) {
+            $self->debug(3, "No dhcp removal with reinstall set for $node");
+        } else {
+            $self->debug (4, "Going to add dhcp entry of $node to remove");
+            $ec = $mgr->new_remove_entry($node);
+        }
+    }
```
stdweird: no else catch all? maybe in debug 3?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -701,13 +678,38 @@ sub fetch_profiles
     return %h;
 }
 
-
+# Initiate the Parallel:ForkManager with requested threads if option is given
+sub init_pm {
```
stdweird: `{` on new line

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +870,29 @@ sub check_protected {
...
 }
 
+sub change_dhcp {
+    my ($self, $method, %nodes) = @_;
+
+    my $dhcpmgr = AII::DHCP->new(
+        logfile => $self->option('logfile'), 
+        dhcpconf => $self->option('dhcpconf'), 
+        cfgfile => $self->option('cfgfile'), 
+        log => $self);
```
stdweird: a logfile and a loginstance?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -935,9 +969,16 @@ sub cmds
...
             }
-
-            $self->$cmd (%nodes);
+            # If needed, do DHCP here
+            if ("$cmd" eq "configure" && !$self->option(NODHCP)){ 
+                $self->change_dhcp(CONFIGURE, %nodes);
+            };
             $self->info (scalar (keys (%nodes)) . " nodes to $cmd");
+            $self->$cmd (%nodes);
+            if ("$cmd" eq "remove" &&  !$self->option(NODHCP)) {
```
stdweird: why modify the dhcp after actual remove (and not before, like with configure?)

aii-core/src/main/perl/Shellfe.pm
kwaegema: The ability for the parent to retrieve data structures is new as of version 0.7.6. We don't strictly need it, but useful for testing/debugging

aii-core/src/main/perl/Shellfe.pm
kwaegema: I didn't want to change the order. It was implemented before in configure: aii -> plugins; remove: plugins -> aii. Don't know if this is by design / needed for the plugins

kwaegema:
remarks addressed. Test fails on jenkins on a component i didn't changed , but not on my laptop.

kwaegema:
Added some more fixes. this now runs on our aii server

aii-core/src/main/perl/DHCP.pm
```diff
@@ -421,8 +421,8 @@ sub new_configure_entry
                 ST_IP_TFTP => $tftpserver,
                 MORE_OPT   => $add_txt,
              } );
-        $self->debug(2, "aii-dhcp: add new entry: host = $host fqdn = $fqdn mac = $mac");
-        $self->debug(3, "aii-dhcp: add new entry: additional opts: $add_txt");
+        $self->debug(2, "AII::DHCP: add new entry: host = $host fqdn = $fqdn mac = $mac");
+        $self->debug(3, "AII::DHCP: add new entry: additional opts: $add_txt");
```
stdweird: add the `host=$host` to this message

aii-core/src/main/perl/DHCP.pm
```diff
@@ -461,11 +461,11 @@ sub read_input
         # get input data
         my $fh = CAF::FileReader->new($filename, log => $self);
         if ($EC->error()) {
-            $self->error("aii-dhcp: configurelist error: " .
+            $self->error("AII::DHCP: configurelist error: " .
```
stdweird: all on one line

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,43 +495,56 @@ sub read_input
...
+    my ($error, $changed) = $self->update_dhcp_config($filename);
+    if ($error) {
+        $self->error("AII::DHCP: failed to update dhcpd configuration $filename");
+        return;
+    }
+
+    # restart dhcpd daemon
+    if ($changed) {
+        if ($self->option('norestart')) {
+            $self->verbose("AII::DHCP: config changed but norestart set!");
```
stdweird: add `$filename`

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,43 +495,56 @@ sub read_input
...
+        } else {
+            my $cmd = $self->option('restartcmd');
+            $self->debug(1, "AII::DHCP: restarting dhcpd daemon using cmd '$cmd'");
```
stdweird: add `config $filename cahnged,` to message

stdweird: and this can be verbose or even info

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,43 +495,56 @@ sub read_input
...
+            $self->restart_daemon([split(/\s+/, $cmd)]);
+        } 
+    } else {
+        $self->verbose("AII::DHCP: no changes to $filename: daemon not restarted");
+    }
+    return 1;
+}
+
+# return true if dhcp config need changes
+sub nodes_to_change {
```
stdweird: `{` on next line

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,43 +495,56 @@ sub read_input
...
-                my $cmd = $self->option('restartcmd');
-                $self->debug(1, "aii-dhcp: restarting dhcpd daemon using cmd '$cmd'");
-                # expects an arayref
-                $self->restart_daemon([split(/\s+/, $cmd)]);
-            } else {
-                $self->verbose("aii-dhcp: no changes to $filename: daemon not restarted");
-            }
-        }
+    if ($self->nodes_to_change()) {
+		return 1 if (!$self->update_and_restart()) ;
```
stdweird: fix indentation

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,43 +495,56 @@ sub read_input
...
     } else {
         $self->debug(1, "aii-nbp: there are no changes to dhcpd configuration to make");
```
stdweird: `aii-nbp`? also this is confusing, message should read something like `no nodes to configure and this no chagnes to dhcp configuration`

aii-core/src/main/perl/DHCP.pm
```diff
@@ -495,43 +495,56 @@ sub read_input
...
+            # expects an arayref
+            $self->restart_daemon([split(/\s+/, $cmd)]);
+        } 
+    } else {
+        $self->verbose("AII::DHCP: no changes to $filename: daemon not restarted");
+    }
+    return 1;
+}
+
+# return true if dhcp config need changes
```
stdweird: that is not what this method does. it only chekcs if there any nodes scheduled for changes. it's true that if no nodes are scheduled, no config change is required. but not the other way around.

aii-core/src/main/perl/Shellfe.pm
stdweird: add this as comment 😉

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -30,8 +30,10 @@ use File::Path qw(mkpath rmtree);
 use File::Basename qw(basename);
 use DB_File;
 use Readonly;
-
+use Parallel::ForkManager 0.7.6;
 our $profiles_info = undef;
+use AII::DHCP;
+use 5.10.1;
```
stdweird: flag this PR as backwards incompatible, this breaks el5 support.

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -229,19 +231,23 @@ sub app_options
          HELP    => 'where to store lock files',
          DEFAULT => "/var/lock/quattor" },
 
+       { NAME    => 'parallel=i',
+         HELP    => 'run commands on hosts in parallel ',
+         DEFAULT => 0 },
+
+        { NAME    => 'logpid',
```
stdweird: fix indent

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -229,19 +231,23 @@ sub app_options
...
 
          # Options for DISCOVERY plug-ins
-         { NAME => 'dhcpconf=s',
-           HELP => 'name of dhcp configuration file',
-           DEFAULT => '/etc/dhcpd.conf' },
-
-         { NAME => 'restartcmd=s',
-           HELP => 'how to restart the dhcp daemon',
-           DEFAULT => 'service dhcpd restart' },
+         { NAME => 'dhcpcfg=s',
```
stdweird: fix indent

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -229,19 +231,23 @@ sub app_options
          HELP    => 'where to store lock files',
          DEFAULT => "/var/lock/quattor" },
 
+       { NAME    => 'parallel=i',
+         HELP    => 'run commands on hosts in parallel ',
+         DEFAULT => 0 },
+
+        { NAME    => 'logpid',
+          HELP    => "Add process ID to the log messages (disabled by default)",
```
stdweird: `by default, unless parallel is > 1` (or somethign like this)

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -298,8 +304,16 @@ sub _initialize
     $self->{LOG_APPEND} = 1;
     $self->{LOG_TSTAMP} = 1;
     $self->{status} = 0;
+
     $self->SUPER::_initialize (@_) or return undef;
-    $self->set_report_logfile ($self->{LOG});
+
+    # Defaults: append to logfile, add timestamp
+    my $logopts = 'at';
+    if ((!defined($self->option('logpid')) && $self->option('parallel') ) || $self->option('logpid'))  {
```
stdweird: only if parallel > 1?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -412,45 +426,11 @@ sub run_plugin
...
-            push(@dhcpop, '--debug', $1);
-        } else {
-            $self->warn("Invalid debug value $dbg passed (should be integer)");
-        };
-    }
-    return @dhcpop;
-}
-
-# Runs aii-dhcp on the configuration object received as argument. It
+# Runs AII::DHCP with the configuration object received as argument. It
```
stdweird: `Runs AII::DHCP` makes no sense

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -412,45 +426,11 @@ sub run_plugin
...
 # uses the MAC of the first card marked with "boot"=true.
 sub dhcp
 {
-    my ($self, $node, $st, $cmd) = @_;
+    my ($self, $node, $st, $cmd, $mgr) = @_;
```
stdweird: can y9ou rename `$mgr` to `$parallel_mgr` or somethig like that? maybe a line of comment that is either undef or an instance of `Parallel`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -464,16 +444,25 @@ sub dhcp
...
     }
 
-    my @commands = $self->dhcpcfg ($node, $cmd, $st, $mac);
-    $self->debug (4, "Going to run: " .join (" ",@commands));
-    my $output;
-    $output = CAF::Process->new(\@commands, log => $self)->output();
-    if ($?) {
-       $self->error("Error when configuring $node: $output");
+    my $ec;
+    if ("$cmd" eq CONFIGURE) {
```
stdweird: why are there`""` around $cmd?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -464,16 +444,25 @@ sub dhcp
...
+            $self->debug(3, "No dhcp removal with reinstall set for $node");
+        } else {
+            $self->debug (4, "Going to add dhcp entry of $node to remove");
+            $ec = $mgr->new_remove_entry($node);
+        }
     } else {
-        $self->debug(4, "Output: $output");
-    };
-    $self->debug(4, "End output");
+        $self->error('dhcp should only run for configure and remove methods');
```
stdweird: add `$node` and esp `$cmd`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -464,16 +444,25 @@ sub dhcp
...
+        $ec = 1;
+    }
+    if ($ec) {
+        $self->error("Error when configuring $node");
```
stdweird: also add `$cmd` to message

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -717,8 +732,25 @@ foreach my $cmd (COMMANDS) {
                     next;
                 };
             };
-            $self->$method($node, $node_states{$node});
+            $self->debug(5, "Going to start $cmd on node $node");
+            if ($pm) {
+                $pm->start($node) and next;
+                $self->verbose("started parallel execution for $node in child");
```
stdweird: please do `fi/esle block, so the `next` case gets logged too

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -701,13 +690,39 @@ sub fetch_profiles
...
+            }
+        );
+        return $pm;
+    } else {
+        return;
+    }
+}
+    
 # Wrapper to execute the commands in sorted manner
 no strict 'refs';
```
stdweird: euuuhhhhhh, what? this is way too much code to disable strict. best to move the `sub {` to its own function, and only disable strict check inside the lofloop and only for the asignment

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +885,31 @@ sub check_protected {
     return %hash;
 }
 
+sub change_dhcp 
+{
+    my ($self, $method, %nodes) = @_;
+    $self->debug(5,"logfile:", $self->option('logfile'), " dhcpcfg:", $self->option('dhcpcfg') );
+    my $dhcpmgr = AII::DHCP->new('script',
```
stdweird: `script`?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +885,31 @@ sub check_protected {
     return %hash;
 }
 
+sub change_dhcp 
+{
+    my ($self, $method, %nodes) = @_;
+    $self->debug(5,"logfile:", $self->option('logfile'), " dhcpcfg:", $self->option('dhcpcfg') );
```
stdweird: space after `,`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +885,31 @@ sub check_protected {
...
+    my $dhcpmgr = AII::DHCP->new('script',
+        "--logfile=".$self->option('logfile'), 
+        "--cfgfile=".$self->option('dhcpcfg'), 
+        log => $self);
+    foreach my $node (sort keys %nodes) {
+        my $st = $nodes{$node}; 
+        $self->debug(3,"Checking dhcp config on node $node for method $method.");
```
stdweird: space after `,`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -412,45 +426,11 @@ sub run_plugin
...
-        } else {
-            $self->warn("Invalid debug value $dbg passed (should be integer)");
-        };
-    }
-    return @dhcpop;
-}
-
-# Runs aii-dhcp on the configuration object received as argument. It
+# Runs AII::DHCP with the configuration object received as argument. It
 # uses the MAC of the first card marked with "boot"=true.
```
stdweird: sort the values that they are looped over (i know, this is stretching the review 😄 )

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +885,31 @@ sub check_protected {
...
+        "--cfgfile=".$self->option('dhcpcfg'), 
+        log => $self);
+    foreach my $node (sort keys %nodes) {
+        my $st = $nodes{$node}; 
+        $self->debug(3,"Checking dhcp config on node $node for method $method.");
+        if ($st->{configuration}->elementExists(DHCPPATH)) {
+            $self->dhcp($node, $st, $method, $dhcpmgr);
+        }
+    }    
+    if ($dhcpmgr->nodes_to_change()) {
```
stdweird: why reinvent the `AII::DHCP->configure` method? or at least fatcor out the common bits

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +885,31 @@ sub check_protected {
...
+         $self->info('DHCP will be updated and restarted');
+         $dhcpmgr->update_and_restart();
+    } else {
+        $self->debug(1, 'DHCP up to date');
```
stdweird: same remark as with `configure`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -935,9 +986,19 @@ sub cmds
             foreach my $node (@reinstall_list) {
                 $nodes{$node}->{reinstall} = 1 if exists($nodes{$node});
             }
-
-            $self->$cmd (%nodes);
+            # If needed, do DHCP here
+            if (!$self->option(NODHCP)){
```
stdweird: better to report this:
if NODHCP:
  verbose or debug (no dhcp)
else:
  ...

stdweird:
@kwaegema nice, only some minor remarks left

aii-core/src/main/perl/Shellfe.pm
kwaegema: @stdweird  ah yes, this is something i still need to figure out. I saw this used in the existing tests. Without this I can't seem to pass the parameters.

aii-core/src/main/perl/Shellfe.pm
kwaegema: aii-dhcp returns 1 on failure of update_and_restart, 0 on success. we don't have this here

kwaegema:
@stdweird done

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -701,24 +691,76 @@ sub fetch_profiles
...
+    my $method = "_$cmd";
+    my %responses; 
+    my $pm = $self->init_pm($cmd, \%responses);
+    foreach my $node (sort keys %node_states) {
+        $self->debug (2, "$cmd: $node");
+        if ($cmd ne 'status') {
+            my $lock = $self->lock_node($node);
+            if (! $lock) {
+                $self->error("aii-shellfe: couldn't acquire lock on $node for $cmd");
+            next;
```
stdweird: fix indent

aii-core/src/main/perl/Shellfe.pm
stdweird: @kwaegema also update the description with was is deprecated.
maybe also mention the change in dhcp update (afaik, dhcp is now also updated before remove; that might also be meaningful)

aii-core/src/main/perl/Shellfe.pm
stdweird: and? this if/else block is common, so please factor it out

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -935,9 +996,21 @@ sub cmds
...
+                # If needed, do DHCP here
+                if ("$cmd" eq "configure"){ 
+                    $self->change_dhcp(CONFIGURE, %nodes);
+                } elsif ("$cmd" eq "remove"){
+                    $self->change_dhcp(REMOVEMETHOD, %nodes);
+                } else {
+                    $self->debug(5, "Method $cmd does not need dhcp");
+                }
+            }
             $self->info (scalar (keys (%nodes)) . " nodes to $cmd");
```
stdweird: this one verbose, the other info is fine (it's almost the same message)

stdweird:
@kwaegema some more minors remarks

kwaegema:
@stdweird done

stdweird:
retest this please

stdweird:
@jrha @ned21 care to review? this PR will conflict with #225, so is best merged asap

aii-core/src/main/perl/Shellfe.pm
jouvin: As for me, we don't need AII to be able to run on a SL5 machine (SL5 is almost EOL if I am right). Only the ability to manage a SL5 node is needed.

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -229,19 +231,23 @@ sub app_options
...
 
          # Options for DISCOVERY plug-ins
-         { NAME => 'dhcpconf=s',
-           HELP => 'name of dhcp configuration file',
-           DEFAULT => '/etc/dhcpd.conf' },
-
-         { NAME => 'restartcmd=s',
-           HELP => 'how to restart the dhcp daemon',
-           DEFAULT => 'service dhcpd restart' },
+       { NAME => 'dhcpcfg=s',
```
jrha: This changes an option name and removes another, breaking backwards compatibility. Please mention this (and any other incompatibilities in the PR description).

jrha:
PR description is completely inadequate, nothing explains WHY this change is being made and none of the backwards incompatibilities are documented.

aii-core/src/main/perl/Shellfe.pm
kwaegema: restartcmd and dhcpconf are options of Shellfe that are used nowhere in the code.. maybe this was copied from aii-dhcp code earlier, but specifying this had never the result you would suspect so it don't really break something

kwaegema:
@jrha updated description and history

aii-core/src/main/perl/Shellfe.pm
jouvin: @kwaegema I missed that but you are wrong... These options are passed to aii-dhcp (or AII:DHCP, I don't remember which one is used) and we rely on them at LAL (we have both a non standard DHCP config file path and a non standard command to restart dhcpd).

jrha:
Much better thanks :+1: it still doesn't explain why we would want to run methods in parallel though (is it faster? :innocent:)

aii-core/src/main/perl/Shellfe.pm
jouvin: In fact, this probably requires moving these options into `/etc/aii/aii-shellfe.conf` instead of `/etc/aii/aii-dhcp.conf`... This should be documented and `ncm-aiiserver` should be fixed to do it for sites that use it to manage their Quattor server (the recommended method IMO).

kwaegema:
@jrha I hope we will achieve faster runs in combination with freeipa
@jouvin what do you mean?

aii-core/src/main/perl/Shellfe.pm
kwaegema: @jouvin these options still exist in the AII::DHCP module, I only removed them from the AII::Shellfe. I don't find where they should have been passed to aii-dhcp nor AII::DHCP. Imho this could only be done by setting this is /etc/aii/aii-dhcp.conf. Can you verify this? 
This is still true for the new code by default. By default AII::DHCP will get his config from   /etc/aii/aii-dhcp.conf unless specified otherwise with the new `dhcpcfg` option.

kwaegema: I checked https://github.com/quattor/configuration-modules-core/blob/master/ncm-aiiserver/src/main/pan/components/aiiserver/schema.pan . the two variables are only valid in structure_aiidhcp, which is only used for generating aii-dhcp.conf. This still exists at https://github.com/quattor/aii/blob/master/aii-core/src/main/perl/DHCP.pm#L58

jouvin: My point is that `AII::DHCP` should not initialize a new `CAF::Application` and I have the feeling it doesn't... So it makes no sense for it to initialize the options and to use a separate config file. `AII::DHCP` should get its options from `/etc/aii/aii-shellfe.conf` as does `aii-pxelinux`. This is why `ncm-aiiserver` should with the same schema file the config files differently...

stdweird:
@jouvin can we move the merge of the configfiles to a later issue/PR? the current PR does not require neither new aiiserver nor new profiles

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -229,19 +231,23 @@ sub app_options
...
+
          # Options for osinstall plug-ins
        { NAME    => 'osinstalldir=s',
          HELP    => 'Directory where Kickstart files will be installed',
          DEFAULT => '/osinstall/ks' },
 
          # Options for DISCOVERY plug-ins
-         { NAME => 'dhcpconf=s',
-           HELP => 'name of dhcp configuration file',
-           DEFAULT => '/etc/dhcpd.conf' },
```
jrha: Hang on... we use this in order to co-exist with other systems!

```
# cat /etc/aii/aii-dhcp.conf 
# File generated by ncm-aiiserver
# Do not edit
dhcpconf = /etc/dhcp/dhcpd.conf.quattor
```

jouvin: It should work as long as the option in your config file is processed... but I am not sure this is the case... we need to ensure that the default is not changed from previous version, else any site relying on the previous default will have a problem...

jouvin:
@stdweird about https://github.com/quattor/aii/pull/239#issuecomment-282010192, I cannot answer. I have no problem if we keep both configuration files for the time being, as long as `AII::DHCP` process `/etc/aii/aii-dhcp.conf`. But see https://github.com/quattor/aii/pull/239#discussion_r102717060: I think (but I may be wrong) that this is not the case as I have the feeling that the CAF::Application is initialized (and config options processed) by `AII::Shellfe`...

aii-core/src/main/perl/Shellfe.pm
stdweird: @jrha it's used by `aii-dhcp` or `AII::DHCP`, this is shellfe.
@jouvin has a valid concern, but we also have this entry, and things still seem to work.

what @kwaegema removed is an option that suggested you could change the dhcpd.conf location via aii-shellfe `.conf` or commanline ; but you can't. aii-dhcp/AII::DHCP however picks up this config file.

aii-core/src/main/perl/Shellfe.pm
stdweird: @jouvin your first sentence is a bit confusing; but i think i get your remark and concern.
however, you can instantiate as many `CAF::Application` as you want, they are separate instances.

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -859,6 +895,27 @@ sub check_protected {
     return %hash;
 }
 
+sub change_dhcp 
+{
+    my ($self, $method, %nodes) = @_;
+    $self->debug(5, "logfile:", $self->option('logfile'), " dhcpcfg:", $self->option('dhcpcfg'));
+    my $dhcpmgr = AII::DHCP->new('script',
+        "--logfile=".$self->option('logfile'), 
+        "--cfgfile=".$self->option('dhcpcfg'),
```
stdweird: @jouvin @jrha this is the important bit. you can also see in the diff that the other options are not part of any removed lines.
maybe the name should be `--aiidhcpcfg` to make it more clear

jouvin: @stdweird thanks for the clarification. I now understand that `AII::DHCP` initialize a different CAF::Application. I found this not very clean but I think we can open an issue for the next release (about unifying config files and have only shared by `aii-shellfe` and all its components).

PR #242, opened 2017-02-09
aii-ks: relax test regexp due to fp rounding issues

stdweird:
Reported during #225 

Tests fail due to `checksize` kickstart code not matching the regexps

Produced kickstart output (incl reporting)
 * the VERB lines are from ncm-lib-blockdevices
 * the value printed in the pre section is floored first, but shouldn't change between perl versions.
```
# 
# [VERB] Going to use 2 conditions: diff, fraction
# [VERB] Diff defined 100, updated min/max 3900 / 4100
# [VERB] Fraction defined 0.001, updated min/max 3996 / 4004
# correct_disksize_MiB /dev/sdb 3996 4003
```

the issue is the following floating rounding:
```
$  perl -e 'use POSIX qw(floor);$a=(1+0.001)*4000;printf("%d %.30f %d\n", $a, $a, floor($a));'
4003 4003.999999999999545252649113535881 4003
```

### DIFF
--- a/aii-ks/src/test/resources/regexps/pre_blocksize_blockdevices
+++ b/aii-ks/src/test/resources/regexps/pre_blocksize_blockdevices
@@ -1,7 +1,7 @@
 Test the blockdevice generation in the pre section with disk with correct size defined
 ---
 ---
-^correct_disksize_MiB /dev/sdb 3996 4004$
+^correct_disksize_MiB /dev/sdb 3996 400[34]$
 ^if \[ \$\? -ne 0 \]; then$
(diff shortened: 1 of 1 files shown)


stdweird:
@jouvin this is cherry-picked, so shouldn't harm your PR

stdweird:
@jrha should help you continue with the build-tools release process

stdweird:
@jouvin hmm, the cherry pick had a different commit id, so i also cleaned up the commit message. you will have to get rid of your  commit in your PR

PR #247, opened 2017-02-24
aii-core: add perl-Parallel-ForkManager to RPM

### LINKED ISSUE #246: aii-core: missing perl-Parallel-ForkManager dependency
#239 makes required to install `perl-Parallel-ForkManager` to be able to run `aii-shellfe`.

jouvin:
Fixes #246

### DIFF
--- a/aii-core/pom.xml
+++ b/aii-core/pom.xml
@@ -207,6 +207,7 @@
             <requires>
               <require>perl-CDB_File</require>
               <require>perl-XML-Simple</require>
+              <require>perl-Parallel-ForkManager</require>
             </requires>
             <mappings>
               <mapping>


aii-core/pom.xml
```diff
@@ -207,6 +207,7 @@
             <requires>
               <require>perl-CDB_File</require>
               <require>perl-XML-Simple</require>
+              <require>perl-Parallel-ForkManager</require>
```
stdweird: why is there any `require` at all? without requires, the perl `use` lines get translated in dependencies.

jouvin: You may be right... In fact the real issue was https://github.com/quattor/template-library-os/pull/90 and I forgot the automatic generation of dependencies when building RPM...

jouvin: Confirmed by RPM inspection...

PR #249, opened 2017-02-24
aii-core: fix typos in Shellfe.pm

### LINKED ISSUE #248: aii-shellfe broken after #239
To test #225 (rebased on #239), I installed the last CAF. Unfortunately, `aii-shellfe` no longer works with the error:
```
Can't locate object method "options" via package "AII::Shellfe" at /usr/lib/perl/AII/Shellfe.pm line 651
```
This happens when defining `CAF::Download::LWP` options. 

Any idea why this happens and what is the workaround... This is a blocking issue for testing and eventually merging #225.

**This is also a showstopper for next release** as #239 has already been merged.

jouvin:
Fixes #248 introduced by #239

### DIFF
--- a/aii-core/src/main/perl/Shellfe.pm
+++ b/aii-core/src/main/perl/Shellfe.pm
@@ -538,7 +538,7 @@ sub nodelist
                 CAFILE() => 'cacert',
                 CADIR() => 'cadir'
                 );
-            my %lwp_opts = map {$lwp_opts_map{$_} => $self->options($_)} grep {$self->options($_)} keys %lwp_opts_map;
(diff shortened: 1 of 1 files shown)


stdweird:
@jouvin odd, we are (or i tought we were) running this...
i'll open a PR with unittests later

jouvin:
I guess you are not! I looked at the possibility of PR that would detect this but was not sure about what was really doable...

stdweird:
i just checked, we run this code, but we use the `dir::` cdburl, so this code path gets ignored; basically not testing these important changes in code.
the PR was initially meant to fix issues we had with freeipa hook (and those were resovled). 

anyway, apologies for this. i'll look for a PR with unittests

Issue #252, opened 2017-02-28
remove aii-opennebula and move all issues to the component


stdweird:
both aii-opennebula and ncm-opennebula now generate the `quattor/aii/opennebula` templates, and only the component is supported.

the subdir can be deleted, and the `aii-opennebula` entry must be removed form the main project pom.xml.

also any leftover issues have to be closed or transferred to the component repo

alvarosimon:
Old  opennebula aii issues are now closed or reassigned to the component, I will open a new PR to remove opennebula aii..

PR #253, opened 2017-02-28
Remove opennebula from aii

alvarosimon:
OpenNebula aii is now assigned to the component

Fix: #252


alvarosimon:
bye bye opennebula aii 😢

PR #271, opened 2017-05-17
Update el7.pan

seraf1:
The "packagesinpost" option can now be overloaded (? =) because "true" value prevents some machines from being installed.


jrha:
Please do not merge your own pull-requests.
This may appear to be a trivial change, but someone should still review it.

seraf1:
Ok, sorry

stdweird:
@seraf1 can you share any details on what went wrong? why can you install packages with kickstart repo and not with quattor configured one? this feels very odd.

PR #277, opened 2018-01-10
Allow LOCALHOST in a ksuserscript URL to substituted.

ajf8:
As is done in aii-pxelinux, so the file can be served by the host
generating the kickstart.

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -566,6 +566,7 @@ sub ksuserscript
 
     return unless $config->elementExists ($path);
     my $url = $config->getElement ($path)->getValue;
+    $url =~ s{LOCALHOST}{LOCALHOST}e;
     $this_app->debug (5, "User defined script to be fetched ",
                       "from $url for path $path");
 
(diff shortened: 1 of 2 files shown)


aii-ks/src/main/perl/ks.pm
```diff
@@ -566,6 +566,9 @@ sub ksuserscript
 
     return unless $config->elementExists ($path);
     my $url = $config->getElement ($path)->getValue;
+    my $server = LOCALHOST;
+    $url =~ s{LOCALHOST}{$server};
```
stdweird: this behaviour should be added to the documentation, at the very least to the comment above the sub
and :+1: for the confusing, altough you could do better with 
```perl
$url =~ s{LOCALHOST}{LOCALHOST}e;
```

aii-ks/src/main/perl/ks.pm
```diff
@@ -566,6 +566,7 @@ sub ksuserscript
 
     return unless $config->elementExists ($path);
     my $url = $config->getElement ($path)->getValue;
+    $url =~ s{LOCALHOST}{LOCALHOST}e;
```
stdweird: my recommnedation to use this form was more sarcasm wrt perl then genuine advise. i'm glad @jrha doesn't find this confusing 😄

jrha: No more confusing than anything else in Perl...

PR #278, opened 2018-03-16
aii-shellfe: add metaconfig command to generate the metaconfig service files

stdweird:


aii-core/src/main/perl/Shellfe.pm
```diff
@@ -196,7 +202,11 @@ sub app_options
          DEFAULT => undef },
 
        { NAME    => "$PROTECTED_OPTION=s",
-         HELP    => 'required when removing/configuring/(re)installing protected nodes otherwise the operation will fail. A node is protected when /system/aii/protected is set, and the value of that path must be passed to this parameter. Consequently only those protected nodes with same confirmation can be modified in a single shellfe command invocation.',
+         HELP    => 'required when removing/configuring/(re)installing protected nodes otherwise ' .
```
ned21: This change seems unrelated and should be its own commit?

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -214,7 +224,8 @@ sub app_options
          DEFAULT => undef },
 
        { NAME    => 'use_fqdn',
-         HELP    => 'Use the fully qualified domain name in the profile name (if specified). Enable it if you use a regular expression with a dot as "host name"',
+         HELP    => 'Use the fully qualified domain name in the profile name (if specified). Enable it if you ' .
```
ned21: Ditto

stdweird:
@ned21 done

PR #283, opened 2018-05-18
ks: support leavebootorder option

stdweird:
The default feature for anaconda is to change the boot order on systems that support it, for ordinary systems this is desirable, but is a problem for automated management systems like Quattor which rely on the system to always boot from network and fall back to localboot if no boot-time action (i.e. re-install) is required.

This adds support for the Anaconda option which disables the modifcation of boot order.


jrha:
@stdweird I've added a description, please check

PR #285, opened 2018-05-21
 ks: add support for pxeboot: restore pxeboot as first boot order

stdweird:


aii-ks/src/main/perl/ks.pm
```diff
@@ -1587,6 +1601,40 @@ EOF
...
+        print <<EOF;
+#
+# restore pxeboot as first boot order in case efibootmgr is around
+#
+efibootmgr=/usr/sbin/efibootmgr
+function join { local IFS=","; echo "\$*"; }
+if [ -x \$efibootmgr ]; then
+    ordertxt=\$(\$efibootmgr |grep BootOrder | sed "s/.*: \\?//")
+    order=(\$(echo \$ordertxt | tr ',' ' '))
+    pxe=(\$(\$efibootmgr -v | grep -ie ' \\(NIC\\|PXE\\|Network\\) ' | sed "s/*\\? .*//;s/^Boot//" | sort ))
```
jrha: For some reason I'm finding this Bash-in-Perl hard to follow. What are you trying to match here?

I haven't got many UEFI systems lying around, but `efibootmgr -v | grep -ie ' \(NIC\|PXE\|Network\) '` doesn't seem to match anything on them, whereas removing the spaces does match something, but...
```
# efibootmgr -v | grep -ie '\(NIC\|PXE\|Network\)'
Boot0007* Onboard NIC	BBS(Network,Onboard NIC,0x0)AMBO
Boot000A  Onboard NIC(IPV4)	PciRoot(0x0)/Pci(0x19,0x0)/MAC(aabbccddeeff,0)/IPv4(0.0.0.0:0<->0.0.0.0:0,0,0)AMBO
Boot000B  Onboard NIC(IPV6)	PciRoot(0x0)/Pci(0x19,0x0)/MAC(aabbccddeeff,0)/IPv6([::]:<->[::]:,0,0)AMBO
```

jrha: FYI that example was with efibootmgr version 0.12.

jrha:
@stdweird I assume this is in production usage at UGent?

stdweird:
@jrha we run with this patch, but do not use the code (ie pxeboot false. the reason is that we can't pxeboot with localboot anymore, and i haven't figured out why not (i think it never worked to begin with)

so it works, but due to other reasons, we don't enable it

stdweird:
@jouvin do you know if pxeboot with localboot and uefi is supposed to work?

jrha:
At some point we switched from `localboot 0` to `localboot -1` to work around some broken firmware... What are you using?

aii-ks/src/main/perl/ks.pm
wpoely86: Can't you simple take the current boot device from `efibootmgr` and set that as the first device in boot order?

At least you will keep the situation in whatever state it was.

stdweird:
test this please

PR #289, opened 2018-07-14
aii-core: add perl-IPC-Run as a required dependency

### LINKED ISSUE #288: aii-installfe: perl-IPC-Run dependency is not defined as required
`aii-server` RPM is missing the dependency `perl-IPC-Run` that is required to run `aii-installfe`.

jouvin:
- Required by aii-installfe

Fixes #288

### DIFF
--- a/aii-core/pom.xml
+++ b/aii-core/pom.xml
@@ -217,6 +217,7 @@
             <url>https://github.com/quattor/aii/tree/master/aii-core</url>
             <requires>
               <require>perl-CDB_File</require>
+              <require>perl-IPC-Run</require>
               <require>perl-XML-Simple</require>
             </requires>
             <mappings>


jouvin:
Not sure what the test failure is about... Something with JUnit. Is it the same as the Jenkins problem already mentioned by @jrha in another context?

stdweird:
setting the requirements by hand should always be avoided.
what happens when you remove all the requirements? (in particular, does the rpm still have the 3 requirements?)

jouvin:
@stdweird I don't know the reason but this dependency is one of those not generated automatically. I tend to remember that the 2 other ones were added for exactly the same reason but I didn't check recently. The reason for the issue and for the PR is that this dependency was not installed by YUM, leading to a failure to run `aii-installfe`. Except if a better fix is found, it would be good to have this PR merged by 18.9.

jrha:
@hpcugentbot please retest this

jrha:
@hpcugentbot retest this please

jouvin:
I'd appreciate to have this in 18.6 if possible... Not sure why @stdweird postponed it to 18.9. I spent again some time trying to understand why PErl dependencies are not properly detected when the RPM is built but I didn't manage to find the culprit. Having a dependency manually set is better than a broken application...

jouvin:
😄 Probably when I wrote this, I had in mind that 18.6 was about to be released...

jrha:
:suspect:

PR #291, opened 2018-07-16
aii-installfe: add support for --cdburl

### LINKED ISSUE #290: 18.6.0: aii-installfe should allow to define cdburl
`aii-installfe` should allow to define `cdburl` parameter in the executed `aii-shellfe`, as it does for most `aii-shelffe` parameters. This would make possible to easily share an AII server between SCDB and Aquilon, allowing the Aquilon broker to pass the `cdburl` matching itself .

jouvin:
- Allow to pass aii-shellfe the cdburl to use: particularly usefull
when used from the Aquilon broker (allow the broker to set the cdburl)

Fixes #290

### DIFF
--- a/aii-core/src/main/perl/aii-installfe
+++ b/aii-core/src/main/perl/aii-installfe
@@ -66,6 +66,16 @@ expressions can be used.
 Select boot from local disk for hosts listed on <filename>. Hosts have to
 be listed one per line. Lines starting with # are comment.
 
+=item --cdburl <url>
+
+URL for CDB location
+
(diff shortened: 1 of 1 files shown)


stdweird:
i've approved the PR, since i've never used it (i didn't even know that it existed).
but i think more options are out of sync, but i guess this is sufficient for your aq work.

jouvin:
@stdweird I didn't take the time to check for other options out of sync. If you identified some, I'm happy to fix them at the same time.

stdweird:
retest this please

PR #292, opened 2018-07-26
ks.pm: do not redirect to /dev/null dd erros

jouvin:
- dd errors may be useful to troubleshoot partition wiping


stdweird:
looks ok but tests fail

jouvin:
Sorry, I missed it was my fault 😄 Should be fixed now.

PR #294, opened 2018-08-24
aii: ks: disable selinux with kernel param

### LINKED ISSUE #293: Disabling selinux during OS installation does not work with CentOS 7.5
Hi,
By default, selinux is disabled during the OS installation.
However, starting with CentOS 7, selinux is only disabled during the installation, but is set as enabled after the installation. It was until now a minor issue, and was probably due to the missing ```selinux-policy``` package. We can see the following line in the ```anaconda.log``` file:
```
ERR anaconda: Error setting selinux mode: [Errno 2] No such file or directory: '/mnt/sysimage/etc/selinux/config'
```

With CentOS 7.5, the issue becomes worse, as the selinux context is not set on the filesystem and it prevent systemd to start correctly. Setting the **AII_OSINSTALL_SELINUX** variable to ```permissive``` makes the installation succeed.
At least two solutions are available :I would suggest not to remove the selinux packages or to set the selinux to permissive by default on CentOS7.
Jerome

stdweird:
Fixes #293


stdweird:
@Pansanel i have not tested this, feel free to try

PR #296, opened 2018-09-05
DHCP plugin updates

gombasg:
Add tests, adapted from aii-core. Replace bare open() with
CAF::FileReader to make the tests work. Include some improvements which
accumulated on our side.


jrha:
Are there any examples of `/system/aii/discovery` being used?

aii-dhcp/src/main/perl/dhcp.pm
```diff
@@ -312,12 +310,24 @@ sub Configure
     }
     my $ip = $self->get_ip($bootable, $tree);
 
+    my $server_ip = gethostbyname(hostname());
+    $server_ip = inet_ntoa($server_ip) if defined($server_ip);
```
stdweird: should be `Socket::inet_ntoa`.

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
```diff
@@ -2,24 +2,31 @@
...
-#"/system/aii/dhcp/addoptions" = "options blu-blo-bli bla;";
-#
+"tftpserver" ?= AII_DHCP_TFTPSERVER;
+
+@documentation{
+    Additional options to include in the host definition
+}
 variable AII_DHCP_ADDOPTIONS ?= null;
-"/system/aii/dhcp/options/addoptions" ?= AII_DHCP_ADDOPTIONS;
+"options" ?= AII_DHCP_ADDOPTIONS;
```
stdweird: shoudl be `"options/addoptions"`

aii-core/src/main/perl/Shellfe.pm
```diff
@@ -537,6 +538,10 @@ sub dhcp
 {
     my ($self, $node, $st, $cmd, $dhcpmgr) = @_;
 
+    # If the profile has a discovery plugin configured, then don't second-guess
+    # it - it may or may not be ISC DHCP
+    return if $st->{configuration}->elementExists (DISCOVERY);
```
stdweird: add a `$self->verbose("discovery plugin configured, no dhcp configuration")` or something like that

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
```diff
@@ -2,24 +2,31 @@
...
-# "/system/aii/dhcp/options/tftpserver" = "tftp.mydomain.org"
-#
+bind "/system/aii/discovery/dhcp" = structure_dhcp_module_info;
+
+prefix "/system/aii/discovery/dhcp";
+
+@documentation{
+    Enable the plugin
+}
+"enabled" = true;
```
stdweird: seriously backwards incompatible
i propose to set `"enabled" ?= false;`

aii-core/src/main/perl/Shellfe.pm
stdweird: add a $self->verbose("discovery plugin configured, no dhcp configuration") or something like that

stdweird: and the return condition should be that `$st->{configuration}->{enabled}` is true (so that one can disable it; and it should be disabled by default)

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
```diff
@@ -2,24 +2,31 @@
...
-include 'quattor/aii/${project.artifactId}/schema';
+include 'quattor/aii/dhcp/schema';
 
-# TFTP server
-# This is optional. Only it is necessary if your TFTP server is running on
-# a different machine than the DHCP server
-#
-# "/system/aii/dhcp/options/tftpserver" = "tftp.mydomain.org"
-#
+bind "/system/aii/discovery/dhcp" = structure_dhcp_module_info;
```
stdweird: fyi, since you do this, there's no way to unbind so the code has to be able to disable the discovery

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
gombasg: No, it should not :-) The original `aii-dhcp` command had a command-line option called `--addoptions` - which, due to being a command line option, needed to be a single string. But the plugin version never had "addoptions".

In other words - originally, the DHCP plugin did not come with templates, which was unfortunate. When templates were eventually added, they did not match what the code did. This change makes the templates match the code.

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
gombasg: I disagree :-) When `quattor/aii/ks/config` is included, it makes sure `/system/aii/osinstall/ks` is not empty, so the plugin will actually be used.

This template does exactly the same. If I write `include 'quattor/aii/dhcp/config'`, I _want_ that to cause the plugin to be enabled. But, since the plugin does not have any plugin-wide configuration options at the moment, the `enabled` flag is used to serve this purpose.

You could call `enabled` as `an_arbitrary_value_to_ensure_the_path_exists`, but that is much harder to type :-)

aii-core/src/main/perl/Shellfe.pm
gombasg: I have no problem with adding a diag message, however, the code above is correct. The use of plugins is triggered by the existence of the plugin name - there does not have to be any specific keys under the name; see `iter_plugins()`. Neither `nbp` nor `osinstall` plugins are required to have a property called `enabled` - I don't see why `discovery` plugins should have that.

As I wrote above, `enabled` is really just a placeholder here - since the DHCP plugin does not have any plugin-wide options, it needs _something_ to force the path to exist.

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
stdweird: euhm, my point is that this bind disables dhcp configuration and only configures discovery; and there's no way to disable it.
so the template with this defined should just be called `dhcp/discovery.pan`, `dhcp/config.pan` at least suggest it does something with dhcp.

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
stdweird: with current code, including `quattor/aii/dhcp/config` sets the discovery plugin, and effectively disables any dhcp configuration. i'm not sure it's the intention, at the very very least is backwards incompatible

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
stdweird: we use the `options/addtoptions` value; at the very least this change is backwards incompatible
same for the relocation of the tftpserver from `options/tftpserver` to simple `tftpserver`.
in the `Shellfe.pm` is a methods called `dhcp` that uses this. please also fix that code if you plan to cleanup the schema.

aii-core/src/main/perl/Shellfe.pm
stdweird: ok, if you really only need/want path existance, get rid of the enabled; the name suggests that you can also disable it.
you can just do 
```
discovery : dict = dict() with length(SELF) == 0
```
and get the same result.

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
gombasg: No, it is not backwards incompatible - _using_ addoptions in the plugin's templates would be backwards incompatible :-)

The underlying issue is, `aii-core/src/main/perl/DHCP.pm` is not compatible with the DHCP plugin - and therefore, you cannot use the the same templates. Which is unfortunate, but this is the case ever since the plugin exists. `aii-dhcp/src/main/pan/quattor` contains templates for the plugin - so it must be compatible with what the plugin is doing.

This PR does _not_ try to address the incompatibilities between the two implementations. But I think it is successfully pointing out where the incompatibilities are :-)

aii-dhcp/src/main/pan/quattor/aii/dhcp/config.pan
gombasg: Well, `ks/config.pan` and `pxelinux/config.pan` are not called `ks/osinstall.pan` and `pxelinux/nbp.pan` either :-)

The current template layout is inconsistent - it mixes core bits and plugins together. Maybe the templates of real plugins should be moved to a `plugins` subdirectory?

stdweird: my point wrt renaming `dhcp/config` to `dhcp/discvovery` is that the current template and code does not configure dhcp since it enables the discovery plugin/whatever.

i'm aware that it's a mess (and i clearly haven't looked at the code base for a while), but this PR breaks our setup, so i'm trying to figure out where it goes wrong and we can get out of this.

stdweird:
@gombasg and @stdweird agreed to rename existing `config.pan` to `embedded.pan`; add the new `config.pan` as `discovery.pan`; add a variable to select between embedded and discovery and let the new copnfig.pan do the proper include

jrha:
Rebased on main

PR #297, opened 2018-09-12
Another bunch of updates

gombasg:
Various smaller changes all over the place.

PR #298, opened 2018-09-12
Run ks-post-reboot as an offline update service

gombasg:
This PR extends #297 with just one commit, but since I cannot easily test this change outside of our environment, I've decided to open a new PR for it.

Background: installing some services, like AFS, or kernel modules which require a kernel newer than what is used by the installer, require doing a reboot after ks-post-reboot - meaning two reboots overall. For many years now, we saved a reboot by running `/etc/rc.d/init.d/ks-post-reboot` inside the `%post` script. This mostly worked well until RH6, but `systemd` is no longer happy to run in a chroot (which is still not critical even if generates some noise), and we'd like to converge to the upstream code. Running ks-post-reboot as an offline update service means normal services are not started, so will not get confused by running on a half-configured/half-working system. The downside is less visibility into the initial configuration process, which can be mitigated by enabling AII remote logging.


aii-ks/src/main/perl/ks.pm
```diff
@@ -610,7 +621,7 @@ sub ksuserscript
 
     print <<EOS;
 pushd /root
-wget --output-document=userscript $url
+wget --timeout 60 --output-document=userscript $url
```
stdweird: i understand that you want to protect against hanging, but if wget times out, should we abort rather than continue?

stdweird:
@gombasg very minor remark; i'm going to give this a try in short while.

aii-ks/src/main/perl/ks.pm
gombasg: That's a good question. In our case, if `userscript` does not run for whatever reason, then the build will fail pretty soon after that anyway.

gombasg: I've added `|| fail ...`

PR #299, opened 2018-09-13
aii-dhcp: support optional hostname ip verification

stdweird:
Depends on #296


aii-dhcp/src/main/pan/quattor/aii/dhcp/schema.pan
```diff
@@ -20,4 +20,6 @@ type structure_dhcp_dhcp_info = {
       the value you specify in templates.
     }
     "options" ? string{}
+    @{Verify hostname in DNS}
+    "verifyhostname" ? boolean
```
gombasg: I think this should be `choice("yes", "no", "strict")` instead of a boolean, where `strict` would cause the component to fail if `gethostbyname()` below fails. Usually, not having a hard dependency on DNS being up-to-date is good, but I can think of a few use cases where it would be useful if AII could send the message: `You need to wait for DNS to get updated, because otherwise the build will fail later anyway`.

jrha: @stdweird any thoughts?

stdweird: wow, this is long time ago.
sure, but perhaps then `choice("yes", "no", "warning")` or something. or `check_only` instead of `warning`.

jrha:
Rebased on `main`.

PR #302, opened 2018-10-31
Disable selinux in PXE boot when requested

wpoely86:


jrha:
Might be worth doing the same for "enforcing"

PR #307, opened 2019-03-19
aii-pxelinux: Give up regex-based interface name validation

gombasg:
The regex become way too complex, and it interacts badly with custom
interface names. If the value of "ksdevice" is not one of the special
values understood by Anaconda, then it should rather be an existing
interface.


stdweird:
@gombasg do we also drop it in ncm-network?

gombasg:
@stdweird Yes, I'd say drop this from `ncm-network`, too - being able to give interfaces arbitrary names can be useful. Unfortunately, upstream `ncm-network` has become incompatible with our fork, so I'm seriously thinking about switching to NetworkManager (at least for the simple host-with-one-interface-only) cases just to be able to get away from ncm-network.

PR #310, opened 2019-05-13
Revert "ks: Drop end_script" and deprecate instead

jrha:
This reverts commit 9bf2ec9103653da92af8ac9baea475aec356a418 which caused too much backward-incompatibility and instead deprecates the field with a fall-back default in the Perl code.

Further PRs will be introduced to remove this in a friendlier way.


jrha:
Tests pass for me locally now (should really enable Travis on this repo).

PR #311, opened 2019-06-17
Fix the handling of multiple plugins of the same type

gombasg:
Try to run the rest of the plugins even if one has issues (e.g., cannot
be loaded). Don't exit after the first success.


stdweird:
@gombasg i would need to reread the code, but you are sure that a later succesful plugin does nto set the poverall state to "succes". a single failure should result in overall failure imho

gombasg:
@stdweird I'm not sure, because I never saw such an issue :-) On the other hand, I did see the second plugin silently ignored when I played with stock upstream code.

stdweird:
@gombasg i checked the code a bit, the success state is not saved, so any error state cannot be overwritten

PR #313, opened 2019-06-17
aii-ks: end_script removal, try #2

gombasg:
Drop all code referencing end_script, but allow it still being set by
templates.


aii-ks/src/main/pan/quattor/aii/ks/variants/el7.pan
```diff
@@ -8,7 +8,7 @@ prefix "/system/aii/osinstall/ks";
 "langsupport" = null;
 "packages_args" = list("--ignoremissing");
 
-"end_script" = "%end";
+"end_script" = "%end";  # TODO: remove - no longer used
```
stdweird: @gombasg can you make an issue that we should remove it later

PR #315, opened 2019-11-04
RHEL8

gombasg:
RHEL 8 support for AII. Only the creation of /etc/systemd/system/system-update.target.wants is really needed, the rest is mostly cosmetic.

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -83,6 +83,7 @@ use constant   KSDIROPT         => 'osinstalldir';
 use constant ANACONDA_VERSION_EL_5_0 => version->new("11.1");
 use constant ANACONDA_VERSION_EL_6_0 => version->new("13.21");
 use constant ANACONDA_VERSION_EL_7_0 => version->new("19.31");
+use constant ANACONDA_VERSION_EL_8_0 => version->new("29.19");
 use constant ANACONDA_VERSION_LOWEST => ANACONDA_VERSION_EL_5_0;
 
 
@@ -423,8 +424,11 @@ sub kscommands
         $ntp_servers = ' --ntpservers=' . join(',', @{$tree->{ntpservers}});
     }
 
+    if ($version < ANACONDA_VERSION_EL_8_0) {
+        print "install\n";
+    }
+
     print <<EOF;
-install
 $installtype
 reboot
 timezone --utc $tree->{timezone}$ntp_servers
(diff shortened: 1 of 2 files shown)


aii-ks/src/main/perl/ks.pm
```diff
@@ -984,12 +988,16 @@ sub ksinstall_rpm
...
+    my $tree = $config->getElement(KS)->getTree;
+    my $version = get_anaconda_version($tree);
+
     # DISABLED_REPOS doesn't exist in 13.1
     my $disabled = [];
     if ( $config->elementExists(DISABLED_REPOS) ) {
         $disabled = $config->getElement(DISABLED_REPOS)->getTree();
     }
-    my $cmd = "yum -c /tmp/aii/yum/yum.conf -y install ";
+    my $packager = $version >= ANACONDA_VERSION_EL_8_0 ? "dnf" : "yum";
```
stdweird: i don't like this one. it's confusing with the yum.conf, and yum as command is fully supported.

aii-ks/src/main/perl/ks.pm
```diff
@@ -1684,6 +1693,7 @@ EOF_reboot_unit
     ln -sf /etc/rc.d/init.d/ks-post-reboot /system-update
 
     # The documentation recommends creating the .wants symlink directly instead of using [Install] and 'systemctl enable'
+    mkdir -p /etc/systemd/system/system-update.target.wants
```
stdweird: i ran into this one as well, was making a PR too :wink:

stdweird:
@gombasg i'm working on a PR too. i have another bunch of code refactor to allow the configuration of the install repos based on the configured repositories in the profile (instead of the current string). we need appstream as addiitonal repo for early authconfig support.

gombasg:
@stdweird You mean converting `/system/aii/osinstall/ks/repo` from a list-of-strings to a list-of-dicts? That would be nice :-) However, the list of repositories used for Kickstart should remain distinct from the repositories used by spma, populating `/system/aii/osinstall/ks/repo` from the spma repositories would not work for us.

If you're working on that, it would also be nice to make `/system/aii/osinstall/ks/installtype` a dictionary, too, and add support for using `liveimg` :-)

stdweird:
@gombasg no, i'm was going to change the code a bit so you can pass a "magic" name (like `@appstream` to the `/system/aii/osinstall/ks/repo` and it would add all spma repos matching that name. i'm not touching any of the clunky lists that should be dicts :smile:

stdweird:
test this please

PR #314, opened 2019-06-17
New ncm lib blockdevices

gombasg:

PR #280, opened 2018-04-26
Support active configuration attribute for plugins

stdweird:
Fixes #209
Requires quattor/ncm-ncd#74 and #203 
Replaces #210

PR #316, opened 2019-11-18
aii-ks: generate kickstart repo entries based on SPMA configuration

stdweird:
Based on #280 and #314 (and https://github.com/gombasg/aii/pull/1) and requires #315

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -992,7 +992,7 @@ sub ksinstall_rpm
         $disabled = $config->getElement(DISABLED_REPOS)->getTree();
     }
     my $packager = $version >= ANACONDA_VERSION_EL_8_0 ? "dnf" : "yum";
-    my $cmd = "$packager -c /tmp/aii/yum/yum.conf -y install ";
+    my $cmd = "$packager -c /tmp/aii/yum/yum.conf -y -x 'kernel*debug*' install ";
 
     $cmd .= " --disablerepo=" . join(",", @$disabled) . " " if @$disabled;
 
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -1333,17 +1333,14 @@ EOF
 }
 
 
-sub yum_setup
+# create repo information with baseurl and proxy settings
+# return hashref with key the repo name
+sub get_repos
 {
-    my ($self, $config) = @_;
+    my ($config) = @_;
 
-    $self->debug(5,"Configuring YUM repositories...");
+    my %res;
 
-    # SPMA_OBSOLETES doesn't exist in 13.1 , assume false by default
-    my $obsoletes = 0;
-    if ( $config->elementExists(SPMA_OBSOLETES) ) {
-        $obsoletes = $config->getElement (SPMA_OBSOLETES)->getTree();
-    }
     my $repos;
     unless ( $config->elementExists(REPO) ) {
       $this_app->error(REPO." not defined in configuration");
@@ -1351,11 +1348,54 @@ sub yum_setup
     }
     $repos = $config->getElement (REPO)->getTree();
 
+    my ($phost, $pport, $ptype) = proxy($config);
+
+    foreach my $repo (@$repos) {
+        if ($ptype && $ptype eq 'reverse') {
+            $repo->{protocols}->[0]->{url} =~ s{://[^/]*}{://$phost:$pport};
+        }
+
+        $repo->{baseurl} = $repo->{protocols}->[0]->{url};
+
+        # mandatory in 16.4 schema
+        #   these values are the default values in the schema
+        $repo->{enabled} = 1 if(! defined($repo->{enabled}));
+        $repo->{gpgcheck} = 0 if(! defined($repo->{gpgcheck}));
+
+        if (! $repo->{proxy} &&
+            $ptype &&
+            $ptype eq 'forward') {
+            $repo->{proxy} = "http://$phost:$pport/";
+        }
+
+        $res{$repo->{name}} = $repo;
+
+    }
+
+    return \%res;
+}
+
+
+sub yum_setup
+{
+    my ($self, $config) = @_;
+
+    $self->debug(5, "Configuring YUM repositories...");
+
+    # SPMA_OBSOLETES doesn't exist in 13.1 , assume false by default
+    my $obsoletes = 0;
+    if ( $config->elementExists(SPMA_OBSOLETES) ) {
+        $obsoletes = $config->getElement (SPMA_OBSOLETES)->getTree();
+    }
+
+    my $repos = get_repos($config);
+    # error reported in get_repos
+    return if ! $repos;
+
     my $extra_yum_opts = {};
     if ( $config->elementExists(SPMA_YUMCONF) ) {
         $extra_yum_opts = $config->getElement (SPMA_YUMCONF)->getTree();
     }
-
 
     print <<EOF;
 mkdir -p /tmp/aii/yum/repos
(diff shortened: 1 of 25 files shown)


aii-ks/src/main/perl/ks.pm
```diff
@@ -985,7 +985,7 @@ sub ksinstall_rpm
         $disabled = $config->getElement(DISABLED_REPOS)->getTree();
     }
     my $packager = $version >= ANACONDA_VERSION_EL_8_0 ? "dnf" : "yum";
-    my $cmd = "$packager -c /tmp/aii/yum/yum.conf -y install ";
+    my $cmd = "$packager -c /tmp/aii/yum/yum.conf -y -x 'kernel*debug*' install ";
```
gombasg: But... Why? It should be `spma`'s job to control what gets installed. Why do you need to add such custom logic here? And why unconditionally? What happens if someone _wants_ to have kernel debuginfo installed?

aii-ks/src/main/perl/ks.pm
```diff
@@ -427,12 +427,31 @@ timezone --utc $tree->{timezone}$ntp_servers
...
         if ($url =~ m/^@(.+)$/) {
+            # find at least one repo with matching name
+            my $pattern = $1;
+            my @matches = grep {m/$pattern/} sort keys %$repos;
+            if (@matches) {
+                foreach my $reponame (@matches) {
+                    print "repo";
+                    foreach my $key (qw(name baseurl proxy includepkgs excludepkgs)) {
+                        my $val = $repos->{$reponame}->{$key};
+                        print " --$key=". (ref($val) eq 'ARRAY' ? join(',', @$val) : $val) if defined($val);
```
gombasg: If I'm not mistaken, this string should be filtered through `proxy_url()` to apply the global proxy settings, because we depend on that instead of having the proxy definitions repeated for every repository.

aii-ks/src/main/perl/ks.pm
```diff
@@ -427,12 +427,31 @@ timezone --utc $tree->{timezone}$ntp_servers
...
 
+    my $repos = get_repos($config);
+    # error reported in get_repos
+    return if ! $repos;
+
     foreach my $url (@{$tree->{repo} || []}) {
         if ($url =~ m/^@(.+)$/) {
+            # find at least one repo with matching name
+            my $pattern = $1;
+            my @matches = grep {m/$pattern/} sort keys %$repos;
```
gombasg: I'd really prefer if the pattern was anchored on both ends by default. I also think interpreting the value as a glob rather than as a regular expression would be simpler to use. I can see being able to write `@productname*` being useful to select all repositories related to a product, but I'd really not want to see templates/repository names which would require more complex matching :-)

aii-ks/src/main/perl/ks.pm
```diff
@@ -1384,29 +1424,19 @@ end_of_yum_conf
...
-        $repo->{gpgcheck} = 0 if(! defined($repo->{gpgcheck}));
+    foreach my $name (sort keys %$repos) {
+        my $repo = $repos->{$name};
 
         print <<EOF;
-[$repo->{name}]
-name=$repo->{name}
-baseurl=$repo->{protocols}->[0]->{url}
+[$name]
+name=$name
```
gombasg: I think this should remain `$repo->{name}`. Nothing requires the name attribute to be the same as the dictionary key was. I don't know if that distinction would matter in practice, though.

aii-ks/src/main/perl/ks.pm
```diff
@@ -408,16 +408,8 @@ sub kscommands
...
-    if ($installtype =~ /http/) {
-        my ($proxyhost, $proxyport, $proxytype) = proxy($config);
-        if ($proxyhost && $proxytype eq "reverse") {
-            if ($proxyport) {
-                $proxyhost .= ":$proxyport";
-            }
-            $installtype =~ s{(https?)://([^/]*)/}{$1://$proxyhost/};
-        }
-    }
+    my $proxy_config = [proxy($config)];
```
gombasg: Passing the information down as a bare list makes it more difficult to follow what's going on. Could you make `sub proxy()` return a hashref instead? Then you could write `$proxy_config->{host}` below, which would look so much nicer :-)

jrha:
Bumping to allow time for discussion.

aii-ks/src/main/perl/ks.pm
stdweird: good idea, i switch to using a glob

aii-ks/src/main/perl/ks.pm
stdweird: no, the `get_repos` call does this already.

aii-ks/src/main/perl/ks.pm
stdweird: done

aii-ks/src/main/perl/ks.pm
stdweird: but it does it badly. also using `proxy_url` there

aii-ks/src/main/perl/ks.pm
stdweird: what dictionary key? there is only one name, repositories in the `/software/repositories` are a list

stdweird:
@gombasg @jrha i'm pushing new functionality/behaviour to this PR

stdweird:
@gombasg @jrha @ned21 this is potentially backwards incompatible by handling the `disabled_repos` differently and more important any other code that uses the aii yum.repo file.

ned21:
Commits  c180663 and  fca0ace seem to have a corrupted commit log? I think @gombasg's comment about [kernel debug packages](https://github.com/quattor/aii/pull/316/files#r357083409) needs to be addressed too?

stdweird:
@ned21 i forgot to squash those 2 commits in previous ones.
the debug remark of @gombasg is now solved by passing the disallowed/ignored package globs also to the yum install command in post.

aii-ks/src/main/perl/ks.pm
stdweird: this is now handled differently (and customisable)

jrha:
@hpcugentbot test this please

aii-ks/src/main/perl/ks.pm
jrha: @gombasg  `get_repos` uses `$repo->{name}` as the key for the hash it returns, so I believe this is doing what you want.

aii-pxelinux/src/main/pan/quattor/aii/pxelinux/schema.pan
```diff
@@ -11,8 +11,14 @@ include 'pan/types';
...
 }
 type structure_pxelinux_pxe_info = {
-    "initrd" : string
+    @{Kernel path (string in exact syntax).
+      If this contains a '@pattern@' substring, the kernel path is generated based on
+        the (first) enabled SPMA repository with name matching this glob pattern (without the '@').}
     "kernel" : string
+    @{Initrd path (string in exact syntax).
+      If this contains a '@pattern@' substring, the initrd path is generated based on
+        the (first) enabled SPMA repository with name matching this glob pattern (without the '@').}
```
jrha, line 20: Is this comment correct? How do the `kernel` and `initrd` paths related to SPMA repository names?

aii-pxelinux/src/main/pan/quattor/aii/pxelinux/schema.pan
```diff
@@ -11,8 +11,14 @@ include 'pan/types';
     PXE configuration
 }
 type structure_pxelinux_pxe_info = {
-    "initrd" : string
+    @{Kernel path (string in exact syntax).
+      If this contains a '@pattern@' substring, the kernel path is generated based on
+        the (first) enabled SPMA repository with name matching this glob pattern (without the '@').}
```
jrha, line 16: Is this comment correct? How do the `kernel` and `initrd` paths related to SPMA repository names?

stdweird:
```
          "initrd": "@rhel_kickstart@/images/pxeboot/initrd.img",
          "kernel": "@rhel_kickstart@/images/pxeboot/vmlinuz",
```

with eg https://download.rockylinux.org/pub/rocky/8/BaseOS/x86_64/kickstart/ the (disabled) kickstart repo.

jrha: Aha, I see, thanks!

jrha (approved):
AFAICT

PR #318, opened 2019-11-22
aii-core: Shellfe: clarify meaning of only_module and restore working of metaconfig option

stdweird:
Was broken after #300


gombasg:
Looks good from a distance, but I need to find some time to check the details.

PR #323, opened 2020-01-15
aii-pxelinux: Fix UEFI boot over http

gombasg:
If the kernel/initrd specified in the profile uses the
"(http,XXXX)/path/to/image" syntax, then the extra slash which is
currently inserted unconditionally at the beginning causes Grub to fail.
Make sure a path separator is inserted only if GRUB2_EFI_KERNEL_ROOT is
in fact not empty.

### DIFF
--- a/aii-pxelinux/src/main/perl/pxelinux.pm
+++ b/aii-pxelinux/src/main/perl/pxelinux.pm
@@ -484,9 +484,17 @@ sub _write_grub2_config
         return 0;
     };
     my $kernel_root = $this_app->option(GRUB2_EFI_KERNEL_ROOT);
-    $kernel_root = '' unless defined($kernel_root);
-    my $kernel_path = "$kernel_root/$pxe_config->{kernel}";
-    my $initrd_path = "$kernel_root/$pxe_config->{initrd}";
+    my $prefix = defined($kernel_root) ? "$kernel_root/" : "";
+
+    # Avoid having an uncoditional "/" at the beginning, because that would
+    # break if the kernel/initrd location uses the "(http,XXXX)/..." or
+    # "(tftp,XXXX)/..." syntax
+    my ($kernel_path, $initrd_path);
+    $kernel_path = $pxe_config->{kernel} =~ m/^\((http|tftp),/ ? "" : $prefix;
+    $initrd_path = $pxe_config->{kernel} =~ m/^\((http|tftp),/ ? "" : $prefix;
+
+    $kernel_path .= $pxe_config->{kernel};
+    $initrd_path .= $pxe_config->{initrd};
 
     $kernel_path =~ s{\bLOCALHOST\b}{LOCALHOST}e;
     $initrd_path =~ s{\bLOCALHOST\b}{LOCALHOST}e;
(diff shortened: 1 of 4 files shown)


gombasg:
@jrha I could not test this change in a live environment yet, but hand-editing the grub configuration and removing the extra slash did work. Hopefully this does not break anything - tftp servers should be fine with paths not starting with "/".

aii-pxelinux/src/main/perl/pxelinux.pm
```diff
@@ -484,9 +484,17 @@ sub _write_grub2_config
...
     my $kernel_root = $this_app->option(GRUB2_EFI_KERNEL_ROOT);
-    $kernel_root = '' unless defined($kernel_root);
-    my $kernel_path = "$kernel_root/$pxe_config->{kernel}";
-    my $initrd_path = "$kernel_root/$pxe_config->{initrd}";
+    my $prefix = defined($kernel_root) ? "$kernel_root/" : "";
+
+    # Avoid having an uncoditional "/" at the beginning, because that would
+    # break if the kernel/initrd location uses the "(http,XXXX)/..." or
+    # "(tftp,XXXX)/..." syntax
+    my ($kernel_path, $initrd_path);
```
stdweird: or 
```perl
my $kernel_path = my $initrd_path = $pxe_config->{kernel} =~ m/^\((http|tftp),/ ? "" : $prefix;
```

ned21: 😨 is how my brain feels when I try to parse that! Too many = operators!

ned21: > or
> 
> ```perl
> my $kernel_path = my $initrd_path = $pxe_config->{kernel} =~ m/^\((http|tftp),/ ? "" : $prefix;
> ```

😨 is how my brain feels when I try to parse that! Too many = operators!

stdweird:
this PR has some conflicts with #316. i can port these patches to that PR, it would cetainly make my life easier

aii-pxelinux/src/main/perl/pxelinux.pm
```diff
@@ -484,9 +484,17 @@ sub _write_grub2_config
...
+    $kernel_path = $pxe_config->{kernel} =~ m/^\((http|tftp),/ ? "" : $prefix;
+    $initrd_path = $pxe_config->{kernel} =~ m/^\((http|tftp),/ ? "" : $prefix;
```
stdweird, line 494: actually, are we sure that this is `$pxe_config->{kernel}` and not `$pxe_config->{initrd}`?

stdweird:
i added in #316 a fix for this

stdweird:
see https://github.com/quattor/aii/pull/316/commits/bc10f4583a139873019799320911ac2848eb92eb

aii-pxelinux/src/main/perl/pxelinux.pm
jrha: @gombasg ?

stdweird:
@jrha @gombasg please close the PR and switch to #316

PR #325, opened 2020-03-24
Use wipefs for RHEL6+

gombasg:
Keep the previous custom wiping function for RHEL5 for a little longer.

This branch depends on https://github.com/quattor/ncm-lib-blockdevices/pull/93 due to removal of the `rereadpt` function.

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -1,9 +1,9 @@
 #${PMpre} NCM::Component::ks${PMpost}
 
 use EDG::WP4::CCM::Path qw (escape unescape);
-use NCM::Filesystem 18.12.0;
+use NCM::Filesystem 19.12.1;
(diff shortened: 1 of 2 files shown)


aii-ks/src/main/perl/ks.pm
```diff
@@ -1,9 +1,9 @@
 #${PMpre} NCM::Component::ks${PMpost}
 
 use EDG::WP4::CCM::Path qw (escape unescape);
-use NCM::Filesystem 18.12.0;
+use NCM::Filesystem 19.12.0;
```
stdweird: you will need to bump this, as it will only be released in next release. to get the unittests to pass, i think latest lease, with subminor 1 should work.

jrha:
Rebased and conflicts resolved.

jrha (dismissed):

stdweird (approved):

PR #326, opened 2020-08-07
Use wipefs for RHEL6+

stdweird:
This is #325 with fixes for unittests


stdweird:
@gombasg i cn also make a PR, but we need open PRs for internal tooling to make aii rpms.

stdweird:
closing this since @jrha fixed #325

PR #327, opened 2021-01-06
aii-ks: Update regex tests to handle wipefs changes in ncm-lib-blockdevices

jrha:
`wipe_metadata` is no longer conditional, nor does it take a size argument.

This was caused by the changes merged in quattor/ncm-lib-blockdevices@7ee1375ee8db9edda1b030d6de259a2d3eae2ae9.


jrha:
@stdweird I've merged this immediately in order to get nightlies building again, but feel free to post-review and revert if you think that this shouldn't be this way.

stdweird:
@jrha euh, sure i guess. there are other PRs opepn to deal with wipefs, so i need to recheck them

PR #329, opened 2021-04-23
Document use of RewriteRules to restrict access

### LINKED ISSUE #111: Idea: securing profiles and kickstart directories
Hi,
The 'profiles' and 'kickstart' directories are currently accessible through HTTP by at least every machine managed by AII. One problem I see is that a machine accessible by regular users could read all configuration information (including passwords, e.g. mysql) of all other machines.

I came up with a simple solution to isolate these files using an automatically-generated .htaccess file. Although not completely secure (regular users are still able to download the particular machine's configuration they are on), at least the information they have access to is much more limited.

Here's a sample secure.sh script I'm testing. I had a talk with @jouvin about it, who suggested me to submit my idea here to get feedback.

```
#!/bin/bash
# Directory containing the profile files
profilesDir='/var/www/html/profiles/'
# File type we're sure to meet in the directory
fileType='json'
for f in `ls $profilesDir/*.$fileType | xargs -n1 basename`; do
    host="${f%.$fileType}"
    h=`host $host | cut -d' ' -f4`
    echo -e "<FilesMatch $host.*>\n\tOrder deny,allow\n\tDeny from all\n\tAllow from $h\n</FilesMatch>"
done
```

I'm calling it this way:

```
secure.sh > /.../profies/.htaccess &
secure.sh > /.../kickstart/.htaccess &
```

There would be some changes to make in order to integrate it with AII, such as updating the .htaccess upon calling "configure" instead of regenerating everything (resolving hostnames could take a lot of time without a local DNS server).

Do you think it would be interesting to implement such idea into AII?

jrha:
to profiles and kickstart files

Resolves #111.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,4 +2,45 @@
 
 AII is the automated installation infrastructure.
 
-More documentation will follow here.
+If you are serving host profiles and kickstart files with Apache httpd, access to can be restricted with the a config based on the following.
+This approach is obviously better than nothing, but is not a substitute for use of x509 certificates and/or Kerberos should you require secrets to be distributed in the profiles.
+
+```apacheconf
+# Use Aliases or set DocumentRoot
+Alias /profiles /var/quattor/web/htdocs/profiles
+Alias /kickstart /var/quattor/web/htdocs/kickstart
+
+<location /profiles>
+  Require all granted
+</location>
+
+HostnameLookups Double
+
+RewriteEngine On
+# The following ACL map can be used to specify hosts which can access all profiles
+RewriteMap ACLmap txt:/var/quattor/web/ACLmap.txt
+
+# Redirect every profile request to HTTPS
+RewriteCond %{HTTPS} !=on
+RewriteRule  ^/profiles/(.*)$ https://%{HTTP_HOST}/profiles/$1 [L]
+
+# Serve only the profile corresponding to the remote host if the host is not part of the ACLmap
+RewriteCond ${ACLmap:%{REMOTE_HOST}|NO} NO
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,4 +2,38 @@
...
+RewriteEngine On
+# The following ACL map can be used to specify hosts which can access all profiles
+RewriteMap ACLmap txt:/var/quattor/web/ACLmap.txt
+
+# Redirect every profile request to HTTPS
+RewriteCond %{HTTPS} !=on
+RewriteRule  ^/profiles/(.*)$ https://%{HTTP_HOST}/profiles/$1 [L]
+
+# Serve only the profile corresponding to the remote host if the host is not part of the ACLmap
+RewriteCond ${ACLmap:%{REMOTE_HOST}|NO} NO
```
stdweird, line 27: if you only need this for the aii server, and it is the same one as the one serving the profiles (like in this example); you can remove this, and use `cdburl = dir:///var/quattor/....` in aii-shellfe.conf

README.md
```diff
@@ -2,4 +2,38 @@
...
+RewriteRule  ^/profiles/.*\.(xml|json)(\.gz)?$ /var/quattor/web/htdocs/profiles/%{REMOTE_HOST}.$1$2 [L]
```
stdweird, line 28: it's better than nothing i guess, but this still allows anyone on a host to fetch the profile of the host. i would at least recommend using certificates or kerberos to fetch the profiles, and adapt the rewriterule accourdingly

README.md
jrha: Good point, will document this too.

README.md
jrha: Will also document that point.

jrha:
Better now @stdweird?

stdweird:
@jrha yes

PR #339, opened 2023-07-13
Refine regex to identify PXE entries in available boot entries

### LINKED ISSUE #338: aii-ks 21.12: regex for identifying PXE boot entries not strict enough
`aii-ks` has a great feature that allows to redefine the UEFI boot order to have the PXE entries first. Identifying the PXE entries relies on a regex as there is no other way to do it. But this regex is not strict enough on some machines, like HP servers, where the UEFI boot entry list contains things like `Boot Menu`, `Network Boot`, `PXE Boot`... that are just referring to LILO menus (clearly not what we want) and not to PXE entries. 

I attach the script 
[boot_pxe_first.sh](https://github.com/quattor/aii/files/11969767/boot_pxe_first.sh.txt) I use to reorder the boot list on an installed system where I used a slightly different regex that works at least on Dell and HP servers.

Michel

jouvin:
Fixes #338


jouvin:
Abandoned after Wouter's suggestion in #338 to just set BootCurrent as the first entry.

PR #340, opened 2023-07-18
UEFI boot order definition: set the device currently booted as the first entry

### LINKED ISSUE #338: aii-ks 21.12: regex for identifying PXE boot entries not strict enough

jouvin:
Fixes #338

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -1787,27 +1787,25 @@ EOF
 # restore pxeboot as first boot order in case efibootmgr is around
 #
 efibootmgr=/usr/sbin/efibootmgr
-function join { local IFS=","; echo "\$*"; }
 if [ -x \$efibootmgr ]; then
-    ordertxt=\$(\$efibootmgr |grep BootOrder | sed "s/.*: \\?//")
-    order=(\$(echo \$ordertxt | tr ',' ' '))
-    pxe=(\$(\$efibootmgr -v | grep -Pi '\\s(NIC|PXE|Network)\\s' | sed "s/*\\? .*//;s/^Boot//" | sort ))
-    neworder=("\${pxe[@]}")
-    for idx in "\${order[@]}"; do
-        add=1
-        for pxeidx in "\${pxe[@]}"; do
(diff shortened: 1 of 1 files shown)


aii-ks/src/main/perl/ks.pm
```diff
@@ -1,4 +1,30 @@
-#${PMpre} NCM::Component::ks${PMpost}
```
stdweird, line 1: please revert this part

aii-ks/src/main/perl/ks.pm
```diff
@@ -17,7 +43,7 @@ our $EC = LC::Exception::Context->new->will_store_all;
 
 our $this_app = $main::this_app;
 # Modules that may be interesting for hooks.
-our @EXPORT_OK = qw (ksuserhooks ksinstall_rpm get_repos replace_repo_glob);
+our @EXPORT_OK = qw (ksuserhooks ksinstall_rpm);
```
stdweird, line 46: why is this being removed? @jouvin is your PR based on latest master?

jouvin: You are right, it based on 21.12.1, I'll fix this.

jouvin:
Should be fixed, thanks @stdweird for noticing. 

@jrha I see that you don't plan to add it into the upcoming release. Please, could I ask you to reconsider this as the current version is broken and lead to painful operations on recent HP hardware... I think the approach suggested by @wpoely86 and implemented here is much safer. @stdweird what do you think?

jrha (approved):

PR #344, opened 2024-04-19
Add capability to set rescue mode bootloader filename

### LINKED ISSUE #343: Adding capability to set bootloader filename by profile
Currently it is missing from aii-dhcp plugin to fetch bootloader filename for DHCP config.

This issue was raised by the gap of VMware ESXi UEFI bootloader, which is capable only to boot the hypervisor, while there might be occasion (during a rescue mode boot), when there's a need for LiveCD boot. That's not doable with the factory bootloader, hence it has to be handled in the DHCP config to use different bootloaders in specific cases.

Solution would be to make AII capable to add 'filename' option to DHCP config by Quattor profile, also handling normal/rescue mode separately.

MS-Daniel-Matolcsi:
This commit adds the feature to aii-dhcp plugin to fetch bootloader filename for DHCP config to be included. Also handles situation if default bootloader may differ from the one would be used in case of rescue mode.

Resolves #343.

### DIFF
--- a/aii-dhcp/src/main/perl/dhcp.pm
+++ b/aii-dhcp/src/main/perl/dhcp.pm
@@ -340,10 +340,16 @@ sub Configure
     if ($opts->{tftpserver}) {
         $tftpserver = $opts->{tftpserver};
     }
+
     if ($opts->{filename}) {
         $filename = $opts->{filename};
         $filename =~ s/BOOTSRVIP/$server_ip/;
     }
+
+    if ($opts->{rescue}) {
+        $filename = $opts->{rescue};
+    }
+
     if ($opts->{options}) {
         foreach my $k (sort keys %{$opts->{options}}) {
             $additional .= "option $k $opts->{options}->{$k};\n";
--- a/aii-core/src/main/perl/Shellfe.pm
+++ b/aii-core/src/main/perl/Shellfe.pm
@@ -580,6 +580,22 @@ sub run_plugin
                 $plug->set_active_config($st->{configuration});
             }
 
+	    my @params;
+	    if ($modulename eq "dhcp") {
+		if ($self->option("rescue") && $method eq CONFIGURE){
+		    $self->debug (4, "Setting params for DHCP plugin and rescue mode");
+		    @params = ($st->{configuration}, "rescue");
+		}
+		else {
+		    $self->debug (4, "Setting params for DHCP plugin and non-rescue mode");
+		    @params = ($st->{configuration}, "none");
+		}
+	    }
+	    else {
+		$self->debug (4, "Setting generic params");
+		@params = ($st->{configuration});
+	    }
+
             # The plugin method has to return success
             my $res = eval { $plug->$method ($st->{configuration}) };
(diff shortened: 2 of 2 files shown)


jouvin:
I don't think you need AII to configure anything in DHCP. It is currently support by AII to boot in rescue mode (can in fact be anything you have configured) by executing "aii-shellfe --rescue node.name". To work, you need to write a Grub2 config file rescue.cfg file similar to localboot.cfg that will be used by the node when booting. rescue.cfg can be something like (not sure it is a working configuration):

---

set default=0
set timeout=2
menuentry "Rescue CentOS 7 (x86_64)" {
    set root=(pxe)
    linuxefi /quattor/rescue-el7_x86_64/vmlinuz ro crashkernel=auto rd.lvm.lv=cl/root rd.lvm.lv=cl/swap rhgb
    initrdefi /quattor/rescue-el7_x86_64/initrd.img
    }
}
----

You can also do something similar for legacy BIOS if you need.

Did I miss something in your usecase?

MS-Daniel-Matolcsi:
We specifically need to use a different bootloader when using rescue mode. This is an edge case ESXi hypervisors' UEFI booting:
- normally using VMware's mboot.efi, but it can't do a chainload or boot a LiveCD image
- hence, in case we have to boot from a Linux LiveCD in case of rescue (eg. due to revise hardware / doing an firmware upgrade), we have to switch to grub-efi eg. That is what covered in my change, if it's given in profile, AII would change 'filename' option in DHCP conf.

aii-dhcp/src/main/perl/dhcp.pm
```diff
@@ -315,9 +320,18 @@ sub Configure
...
     my $additional = "";
+    my filename = "";
     if ($opts->{tftpserver}) {
         $tftpserver = $opts->{tftpserver};
     }
+    
+    if ($opts->(filename)) {
+        $filename = $opts->(filename);
+    }
+    if ($rescue eq 'rescue' && $opts->(rescue)) {
```
jouvin, line 331: Why it is not an `elsif`? It makes no point to set the filename to `$opts->(filename)` in rescue mode or did I miss something?

MS-Daniel-Matolcsi: There might be a case when there's 'filename' set, and 'rescue' not, also if both 'filename' and 'rescue' are set.
I wanted to make sure if there's a 'filename' set, get it anyway, and if there's 'rescue' there as well, overwrite it.

stdweird:
@MS-Daniel-Matolcsi is this the same as #357 ?

stdweird:
@MS-Daniel-Matolcsi yours is better, but you will also need the https://github.com/quattor/aii/pull/357/files#diff-bf9d1e359bea07dacdbbcf6dab94b9a2f4c42b8a8fde532bf7e96b3f3d71dbf2R11 line for your PR

stdweird:
https://github.com/quattor/aii/pull/299 also has the filename set option already (and i forgot about it, it's open for soooo long ;)

PR #345, opened 2024-08-05
aii/ks: support for sending mail in el9 using correct s-nail format

### LINKED ISSUE #342: support for  s-nail (mailx replacement) syntax on el9
mailx pkg is no loner available on rhel9. its been replaced with s-nail, however its syntax is different. 
therefore following syntax is no longer valid which is generated by aii.

`env MAILRC=/dev/null from=root\@$fqdn $mailx_smtp mailx -s "\\\$subject" $rootmail`

as a result failure email does not work.  provide support for new format.

quick test check on cli suggests following may work 

`env MAILRC=/dev/null  mailx -:/ -Sfrom=root@$fqdn -Smta=smtp://<smtp_address>:<port> -s "Boot record for `uname -n`"   $rootmail`

aka7:
Current mailx options used in aii doesn't seem to send mail when s-nail is used, provide correct options when s-nail binary exists.

Fixes #342

### DIFF
--- a/aii-ks/src/main/perl/ks.pm
+++ b/aii-ks/src/main/perl/ks.pm
@@ -1079,6 +1079,9 @@ sub kspostreboot_header
 
     my $mailx_smtp = "";
     $mailx_smtp .= " smtp=$tree->{mail}->{smtp}" if $tree->{mail}->{smtp};
+    # support for s-nail on el9
+    my $snail_smtp = "";
+    $snail_smtp .= "  -Smta=smtp://$tree->{mail}->{smtp}" if $tree->{mail}->{smtp};
 
     print <<EOF;
 #!/bin/bash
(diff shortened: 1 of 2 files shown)


aka7:
my attempt to get failure emails to work on el9. would be get your thoughts and tested by someone @jrha @wpoely86 @stdweird

aii-ks/src/main/perl/ks.pm
```diff
@@ -1132,7 +1139,11 @@ success() {
 
     subject="[\\`date +'%x %R %z'\\`] Quattor installation on $fqdn succeeded"
     if [ -x /usr/bin/mailx ]; then
-        env MAILRC=/dev/null from=root\@$fqdn $mailx_smtp mailx -s "\\\$subject" $rootmail <<End_of_mailx
+        mailx_options="from=root\@$fqdn $mailx_smtp mailx"
```
ned21, line 1142: There's now considerable duplication here and no real difference between this and lines 1105-1109. Perhaps it's time to factor this out into a function called send_email(subject)?

aka7: I understand what you mean it's 3 extra line of duplication in fail and success function.
And it's not just the subject, body is different from success and failure. given the benefit I'm not sure its worth the effort.

ned21: It's not so much the extra lines but the sheer complexity of these lines now -- there are benefits of abstracting that out to make the code easier to follow too and make is easier to change again in the future. You could also encapsulate all the mail config into one place, right now it's defined in multiple places (e.g. 1080-1084).

Having said that, this sort of low level embedded scripting has its own complexity of needing to be generated on the fly from within another programme--getting the escaping right is a nightmare so I appreciate if you don't want to add further levels of indirection.

I'll defer to @jrha  and @stdweird as to which they prefer.

jrha: Definitely pros and cons either way, but I don't think it's worth further refactoring at this point.

jrha (approved):

jrha:
Rebased.

jrha:
Tests need updating with new command format, will have a look.

jrha (approved):

PR #346, opened 2024-09-04
workflows: Switch CI action over to our own container image

jrha:
Based on quattor/configuration-modules-core#1706 with additional packages required for AII tests added back in.


jrha:
I'm unsure whether the extra dependencies should be part of the base image or not.

wpoely86 (approved):

wpoely86:
> I'm unsure whether the extra dependencies should be part of the base image or not.

I would keep them here. It's nicer in design and fat images are not really nice for unit tests.

PR #350, opened 2024-11-11
Fixes for EL9 support

jouvin:
Fixes problems detected during [24.10.0-rc4](https://github.com/quattor/release/issues/365)

PR #352, opened 2024-11-12
Define default installation boot protocol as static

jouvin:
- Used to be the case in EL7 but defined in the EL7 variant and lost afterwards


jrha (approved):

jouvin:
`indent` and `panlint` errors are probably the same as in #350. I'll fix them there and rebase once it is merged.

